PR #27, opened 2014-08-03
29098: Send JSON success for customize_save and allow response to be filtered

westonruter:
https://core.trac.wordpress.org/ticket/29098


westonruter:
Merged!

PR #59, opened 2015-01-02
30737: JS templates for Customizer Panels and Sections

westonruter:
https://core.trac.wordpress.org/ticket/30737

### DIFF
--- a/src/wp-admin/js/customize-controls.js
+++ b/src/wp-admin/js/customize-controls.js
@@ -149,6 +149,7 @@
 	Container = api.Class.extend({
 		defaultActiveArguments: { duration: 'fast', completeCallback: $.noop },
 		defaultExpandedArguments: { duration: 'fast', completeCallback: $.noop },
+		containerType: 'container',
 
 		/**
 		 * @since 4.1.0
@@ -161,7 +162,11 @@
 			container.id = id;
 			container.params = {};
 			$.extend( container, options || {} );
+			container.templateSelector = 'customize-' + container.containerType + '-' + container.params.type;
 			container.container = $( container.params.content );
+			if ( 0 === container.container.length ) {
+				container.container = $( container.getContainer() );
+			}
 
 			container.deferred = {
 				embedded: new $.Deferred()
@@ -359,7 +364,26 @@
 		 * Bring the container into view and then expand this and bring it into view
 		 * @param {Object} [params]
 		 */
-		focus: focus
+		focus: focus,
+
+		/**
+		 * Return the container html, generated from its JS template, if it exists.
+		 *
+		 * @since 4.2.0
+		 */
+		getContainer: function () {
+			var template,
+				container = this;
+
+			if ( 0 !== $( '#tmpl-' + container.templateSelector ).length ) {
+				template = wp.template( container.templateSelector );
+				if ( template && container.container ) {
+					return template( container.params );
+				}
+			}
+
+			return '<li></li>';
+		}
 	});
 
 	/**
@@ -369,6 +393,7 @@
 	 * @augments wp.customize.Class
 	 */
 	api.Section = Container.extend({
+		containerType: 'section',
 
 		/**
 		 * @since 4.1.0
@@ -886,6 +911,8 @@
 	 * @augments wp.customize.Class
 	 */
 	api.Panel = Container.extend({
+		containerType: 'panel',
+
 		/**
 		 * @since 4.1.0
 		 *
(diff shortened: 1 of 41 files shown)


src/wp-admin/js/customize-controls.js
```diff
@@ -704,6 +732,43 @@
...
+/*			var template, // @todo dynamic panel containers
+				section = this;
+
+			if ( 0 !== $( '#tmpl-customize-panel-' + section.params.type ).length ) {
+				template = wp.template( 'customize-panel-' + section.params.type );
+				if ( template && section.container ) {
+					return template( section.params );
+				}
+			}
+*/		},
```
westonruter: Commented-out code needs to be removed or implemented.

src/wp-admin/js/customize-controls.js
```diff
@@ -517,6 +528,23 @@
...
+		/**
+		 * Return the section contents, generated from its JS template, if it exists.
+		 *
+		 * @since 4.2.0
+		 */
+		getContainer: function () {
+			var template,
+				section = this;
+
+			if ( 0 !== $( '#tmpl-customize-section-' + section.params.type ).length ) {
```
westonruter: If we take the same approach as `control.templateSelector` then the `Control.renderContent` then I think we could use the same `getContainer` method for both sections and panels.

src/wp-admin/js/customize-controls.js
```diff
@@ -348,7 +351,15 @@
...
 		 */
-		focus: focus
+		focus: focus,
+		
+		/**
+		 * To override in a subclass, return the container's container element generated from the template.
+		 * @abstract
+		 */
+		getContainer: function () {
+			throw new Error( 'Must override with subclass.' );
```
westonruter: Per below, I think we can have just one generic implementation for both panels and sections if we use a `templateSelector` property.

src/wp-includes/class-wp-customize-control.php
```diff
@@ -515,11 +519,11 @@ protected function render_content() {
...
-	        ?>
-	        <script type="text/html" id="tmpl-customize-control-<?php echo $this->type; ?>-content">
-	                <?php $this->content_template(); ?>
-	        </script>
-	        <?php
+        ?>
+        <script type="text/html" id="tmpl-customize-control-<?php echo $this->type; ?>-content">
+                <?php $this->content_template(); ?>
+        </script>
+        <?php
```
westonruter: Needs to be indented with tabs here.

src/wp-includes/class-wp-customize-panel.php
```diff
@@ -302,25 +306,58 @@ protected function render() {
...
+	 * and be override-able.
+	 *
+	 * @since 4.2.0
+	 */
+	final public function print_template() {
+        ?>
+        <script type="text/html" id="tmpl-customize-panel-<?php echo $this->type; ?>-content">
+			<?php $this->content_template(); ?>
+        </script>
+        <?php
```
westonruter: Tab indentation needed.

src/wp-admin/js/customize-controls.js
```diff
@@ -348,7 +351,15 @@
 		 * Bring the container into view and then expand this and bring it into view
 		 * @param {Object} [params]
 		 */
-		focus: focus
+		focus: focus,
+
```
westonruter: Remove whitespace on line.

src/wp-admin/js/customize-controls.js
```diff
@@ -517,6 +528,23 @@
...
+		getContainer: function () {
+			var template,
+				section = this;
+
+			if ( 0 !== $( '#tmpl-customize-section-' + section.params.type ).length ) {
+				template = wp.template( 'customize-section-' + section.params.type );
+				if ( template && section.container ) {
+					return template( section.params );
+				}
+			}
```
westonruter: This method is not always returning a value.

src/wp-admin/js/customize-controls.js
```diff
@@ -162,6 +162,9 @@
 			container.params = {};
 			$.extend( container, options || {} );
 			container.container = $( container.params.content );
+			if ( 0 === container.container.length ) {
+				container.container = $( container.getContainer() );
```
westonruter: I suppose we should rename `Container` to `Group` :smile:

src/wp-admin/js/customize-controls.js
westonruter: Right now would it just be `return $( '#customize-theme-controls > ul' );`? See the `Panel.embed` method above.

src/wp-admin/js/customize-controls.js
```diff
@@ -553,6 +581,7 @@
```
westonruter: We can eliminate `parentContainer` in favor of `panel.getContainer()` per below.

westonruter:
Close in favor of rebased https://github.com/xwp/wordpress-develop/pull/76

PR #61, opened 2015-01-06
Customize Transactions

westonruter:
***README: [Customizer Transactions Proposal](https://make.wordpress.org/core/2015/01/26/customizer-transactions-proposal/)***

Next steps:

- [x] Merge in https://github.com/xwp/wordpress-develop/pull/62 after committed to Core
- [x] Merge in https://github.com/xwp/wordpress-develop/pull/67 after committed to Core
- [ ] Break out phpDoc and code style improvements into separate patch

Trac tickets addressed by (or related to) this PR:

* [#30937](https://core.trac.wordpress.org/ticket/30937): Add Customizer transactions. **Read writeup here.**
* [#30028](https://core.trac.wordpress.org/ticket/30028): Load Customizer preview iframe with natural URL
* [#30936](https://core.trac.wordpress.org/ticket/30936): Dynamically create WP_Customize_Settings for settings created on JS client
* [#27355](https://core.trac.wordpress.org/ticket/27355): Customizer: Add framework for partial preview refreshes
* [#29988](https://core.trac.wordpress.org/ticket/29988): Twenty Fifteen: Use JS/postMessage to update the color scheme instead of triggering a page refresh
* [#20714](https://core.trac.wordpress.org/ticket/20714): Theme customizer: Impossible to preview a search results page
* [#23225](https://core.trac.wordpress.org/ticket/23225): Customizer is Incompatible with jQuery UI Tabs.
* [#29098](https://core.trac.wordpress.org/ticket/29098): Send JSON success for customize_save and allow response to be filtered
* [#30448](https://core.trac.wordpress.org/ticket/30448): Customizer Manager/Settings classes do not provide public save method to use outside of customize.php
* [#28721](https://core.trac.wordpress.org/ticket/28721): Scheduled changes for the customizer
* [#29932](https://core.trac.wordpress.org/ticket/29932): There is no error reporting in the Customizer. (Whenever a setting is changed, the transaction gets updated. This transaction update Ajax request can send back information about failures, including PHP-based sanitization of values beyond sanitization logic copied to JS.)
* [#28536](https://core.trac.wordpress.org/ticket/28536): Add browser history and deep linking for navigation in Customizer preview. (This is related as we can now deep-link to Customizer previews with settings from a transaction applied persistently.)
* [#28602](https://core.trac.wordpress.org/ticket/28602): Transparently route frontend browsing through Customizer for logged-in editors. (With partial refresh realized, we can load Customizer Controls into the frontend itself. And with transactions, we can safely reload the entire page to persist the state.)
* Merge [#30988](https://core.trac.wordpress.org/ticket/30988) for the new preview logic and `WP_Customize_Setting::value()` usage.
* [#29316](https://core.trac.wordpress.org/ticket/29316) The `WP_Customize_Setting::preview()` method should also by default add a `priority>10` filter for the `customize_value_*` filter there, but this is dependent on having `apply_filters( "customize_value_{$this->id}", $this->default, $this);` or `apply_filters( "customize_value_{$this->type}", $this->default, $this);` in place, so that we can add this filter. Without this, you will not get the previewed value for a setting when you call `WP_Customize_Setting::value()` on a custom type.

# TODO

- [x] Allow settings to be dynamically created at runtime.
- [ ] No need for a login failure message because not ajax preview. The heartbeat will continually push the iframe's current URL and the current nonces.
- [ ] Don't allow links with `target=_blank` or `target=_top` attrs?
- [ ] Must include partial refresh as key aspect of redactor since full refresh is not seamless
- [x] Provide way of restoring a draft customizer transaction
- [ ] Lay groundwork for inline partial refresh controls: autoloading Customizer on every logged-in page. By eliminating iframe entirely we don't need the refresh transport at all. Customizer can be loaded on every page.
- [ ] Customizer preview should have a heartbeat and if not sent present a link to go home. Heartbeat is what will alert a user if they disconnect from Customizer and provide a URL to go back. If logged in always load the customizer preview JS so that heartbeat can run.
- [x] Introduce customize transaction class for managing saving of settings
- [x] Every customizer transaction is a draft until saved, when it becomes published. Future dated is possible too.
- [x] When viewing preview outside of iframe, allow admin bar to appear
- [x] Anyone can now preview. Only cap is to update and save
- [x] We need a uuid for the transaction generated upon loading. And update URL with ID once first update. 
- [ ] When loading customizer for first time the transaction ID needs to be added via replaceState after first update.
- [ ] After saving a transaction, return a new transaction UUID for staring a new transaction.
- [ ] Prevent prevent modifying published transactions. As soon as it is published, it becomes locked. If a published transaction is loaded in the Customizer, then its settings should be copied into a new transaction.
- [ ] If a transaction is sharer, then it should have a private status and so not auto deleted. Draft transactions should be deleted if not modified since X. 
- [x] If someone has a link to a customizer transaction then any settings should apply even if no cap to update with new ones.
- [x] Eliminate preview nonces. We only need update and save nonces.
- [ ] We don't need Messenger channels anymore, do we?
- [x] Debounce updating a transaction.
- [x] Merge https://core.trac.wordpress.org/ticket/29098
- [x] Persist query vars if transaction UUID supplied
- [ ] When the transaction is updated we can send back the php sanitized value into js
- [ ] Send no cache headers and robots blocking if transaction uuid supplies
- [ ] Eliminate separate `theme` query param. This can be included with the transaction. See https://core.trac.wordpress.org/ticket/22880#comment:6
- [x] Add support for publishing transaction posts, which causes the settings to `save()`
- [ ] When admin bar is shown when customizing, there should only be one node, one to open the page in the `customize.php` with the transaction loaded.
- [ ] If the user is not authorized when landing on `customize.php` and there is a UUID supplied, then redirect to the permalink (frontend) instead?
- [ ] If a `wp_transaction` post gets created without a `post_name` supplied, we need to force it to be a new UUID.
- [x] Add previewing for Ajax requests.
- [ ] Add permalink button and button to share the preview with someone else.
- [x] Allow preview to be viewed without being logged in.
- [ ] If a setting gets deleted on the client, then `null` should be sent for that setting ID when the transaction is updated so it can be purged from the drafted transaction post.
- [x] Does the json key/value blob in post_content work with unfiltered_html = false? (from @ocean90)
- [ ] Investigate replacing UUID with using a regular `wp_transaction` post ID. This would require creating an empty `wp_transaction` post whenever opening the Customizer, so that the Customizer preview URL can have this ID added to a `customize_transaction_id` query param. Then when `location.reload()` is called in the preview, this ID will be retained and will load the transaction for previewing. Potentially use `history.replaceState()` in the preview frame, though this would not ensure that links in the preview also get the ID.

# Regressions
- [ ] There is a flash of unloaded content now when a preview window is reloaded. This is because `document.write()` is no longer used to replace the `iframe` window's document. Overall this is a good thing because it means that the Customizer preview behaves like a normal request, fixing several side-effects of the old method. But the flash of unloaded content when the preview is reloaded is unfortunate. The solution to this is to implement partial refresh, which will not only eliminate the flash of unloaded content but it will also drastically improve the speed of previewing non-postMessage changes.

# WP-CLI

In its current state, you can create a scheduled transaction post to apply settings at a specific date, though it isn't very user-friendly:

```sh
wp post create \
    --post_type=wp_transaction \
    --post_name=$(cat /proc/sys/kernel/random/uuid) \
    --post_date_gmt="2016-01-01 00:00:00" \
    --post_status="future" \
    --post_content='{"blogname":"Welcome To The Future"}'
```

# REST API

- [ ] Provide REST API endpoints for mutating transactions.
- [ ] Ensure that REST API calls may include `customize_transaction_uuid` to apply setting previews on API responses. 

# Future Ideas

- [ ] You should be able to pop-out the preview iframe into a separate window, and then pop it back in. Likewise, you should be able to pop-open the Customizer pane when on any given frontend page.


src/wp-includes/class-wp-customize-manager.php
```diff
@@ -376,46 +566,38 @@ public function wp_loaded() {
...
-		if ( ! isset( $this->_post_values ) ) {
-			if ( isset( $_POST['customized'] ) )
-				$this->_post_values = json_decode( wp_unslash( $_POST['customized'] ), true );
-			else
-				$this->_post_values = false;
-		}
-
-		if ( isset( $this->_post_values[ $setting->id ] ) )
-			return $setting->sanitize( $this->_post_values[ $setting->id ] );
+		_deprecated_function( __FUNCTION__, '0.4.2', 'WP_Customize_Manager::transaction::get()' );
```
ocean90: Shouldn't `__FUNCTION__ ` be  `__METHOD__`?

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -459,18 +662,18 @@ public function customize_preview_init() {
 	 * @access public
 	 */
 	public function customize_preview_override_404_status() {
-		if ( is_404() ) {
-			status_header( 200 );
-		}
+		_doing_it_wrong( __FUNCTION__, 'No longer used', '4.2.0' );
```
ocean90: Should be  `_deprecated_function( __METHOD__, …`

src/wp-includes/class-wp-customize-manager.php
westonruter: Thx. Fixed inb 81978a9.

src/wp-includes/class-wp-customize-manager.php
westonruter: Thx. Fixed in 81978a9.

src/wp-includes/class-wp-customize-transaction.php
```diff
@@ -0,0 +1,311 @@
...
+	public function save( $status = 'draft' ) {
+
+		$options = 0;
+		if ( defined( 'JSON_UNESCAPED_SLASHES' ) ) {
+			$options |= JSON_UNESCAPED_SLASHES;
+		}
+		if ( defined( 'JSON_PRETTY_PRINT' ) ) {
+			$options |= JSON_PRETTY_PRINT;
+		}
+		$post_content = wp_json_encode( $this->data, $options );
```
westonruter: See Slack discussion on encoding this with `base64_encode` so that we can get around `unfiltered_html` and slashing problems: https://wordpress.slack.com/archives/core-customize/p1422044969000136

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -704,16 +1113,72 @@ public function add_setting( $id, $args = array() ) {
...
+	 *
+	 * @todo $customized should store more than just key/value, but also serialized settings. The update_transaction call should include the setting configs.
+	 *
+	 * @param array $customized mapping of settings IDs to values
+	 * @return WP_Customize_Setting[]
+	 */
+	public function add_dynamic_settings( $customized ) {
+		$new_settings = array();
+		foreach ( $customized as $setting_id => $value ) {
+			if ( isset( $this->settings[ $setting_id ] ) || $this->get_setting( $setting_id ) ) {
```
ocean90: Isn't this the same?

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -704,16 +1113,72 @@ public function add_setting( $id, $args = array() ) {
...
+			 *
+			 * @param string $class
+			 * @param string $setting_id
+			 */
+			$setting_class = apply_filters( 'customize_dynamic_setting_class', $setting_class, $setting_id );
+
+			/**
+			 * Filter a dynamic setting's constructor args.
+			 *
+			 * This filter must return an array, overriding the false default, to be
```
ocean90: …to be what?

src/wp-includes/class-wp-customize-manager.php
westonruter: Right you are. Fixed in 0924db555d745495704febaf050a0ddf2a574caa.

src/wp-includes/class-wp-customize-manager.php
westonruter: Oops. Fleshed out in 939631c381d60bf850c0d0360908d3a34c9d0bf5

src/wp-includes/class-wp-customize-transaction.php
```diff
@@ -0,0 +1,355 @@
...
+ *
+ * Implements transactions for Customizer settings
+ *
+ * @package WordPress
+ * @subpackage Customize
+ * @since 4.2.0
+ */
+class WP_Customize_Transaction {
+
+	const POST_TYPE = 'wp_transaction';
```
westonruter: Consider: `customize_transact`

via @nacin: https://wordpress.slack.com/archives/core-customize/p1423086583000446

src/wp-includes/class-wp-customize-transaction.php
```diff
@@ -0,0 +1,355 @@
...
+		 */
+		$post_content = base64_encode( $post_content );
+
+		if ( ! $this->post ) {
+			$postarr = array(
+				'post_type' => self::POST_TYPE,
+				'post_name' => $this->uuid,
+				'post_status' => $status,
+				'post_author' => get_current_user_id(),
+				'post_content' => $post_content,
```
westonruter: Use `post_content_filtered` instead. This will eliminate the need to base64-encode. 

Via @nacin https://wordpress.slack.com/archives/core-customize/p1423086534000438

src/wp-includes/link-template.php
```diff
@@ -1164,25 +1166,39 @@ function get_post_type_archive_feed_link( $post_type, $feed = '' ) {
...
+
+	$edit_post_link = admin_url( $post_type_object->_edit_link );
+
+	// Inject post ID into URL if called for
+	if ( false !== strpos( $edit_post_link, '%d' ) ) {
+		$edit_post_link = sprintf( $edit_post_link, $post->ID );
+	}
+
+	if ( 'wp_transaction' === $post->post_type ) {
+		$edit_post_link = add_query_arg( array( 'customize_transaction_uuid' => $post->post_name ), $edit_post_link );
```
westonruter: From @nacin:

> I think I may prefer to see the stuff in link-template.php be added by a filter, but I don't know if we always load the manager in the contexts we need it.
> Shoehorning stuff into `get_edit_post_link()` and what looks like `get_permalink()` feels like... shoehorning.
> Minor note, and can go either way.

https://wordpress.slack.com/archives/core-customize/p1423086466000433

src/wp-includes/class-wp-customize-transaction.php
```diff
@@ -0,0 +1,355 @@
...
+	}
+
+	/**
+	 * Generate a customizer transaction uuid
+	 *
+	 * @since 4.2.0
+	 *
+	 * @return string
+	 */
+	static public function generate_uuid() {
```
westonruter: From @nacin:

> generate_uuid() is insane. :simple_smile: I wonder if our own rand functions may be better.
> But no real complaint there, just still reviewing.

https://wordpress.slack.com/archives/core-customize/p1423086302000427

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -560,21 +866,58 @@ public function customize_preview_settings() {
 	/**
 	 * Prints a signature so we can ensure the Customizer was properly executed.
 	 *
+	 * @deprecated
+	 *
 	 * @since 3.4.0
 	 */
 	public function customize_preview_signature() {
-		echo 'WP_CUSTOMIZER_SIGNATURE';
+		_doing_it_wrong( __FUNCTION__, 'Function no longer used.', '4.2.0' );
```
westonruter: For why this is, see discussion with @nacin: https://wordpress.slack.com/archives/core-customize/p1423085807000392

src/wp-admin/js/customize-controls.js
```diff
@@ -36,6 +64,56 @@
...
+
+		request = _.debounce( function () {
+			var req, dirtyCustomized;
+
+			dirtyCustomized = {};
+			api.each( function ( value, key ) {
+				if ( value._dirty ) {
+					dirtyCustomized[ key ] = value();
+				}
+			} );
```
westonruter: Add `api.trigger( 'updateTransaction', dirtyCustomized )` to allow the value to be manipulated?

src/wp-admin/js/customize-controls.js
```diff
@@ -36,6 +64,56 @@
...
+
+			req = wp.ajax.post( 'customize_update_transaction', {
+				nonce: api.settings.nonce.update,
+				wp_customize: 'on',
+				customized: JSON.stringify( dirtyCustomized ),
+				customize_transaction_uuid: api.settings.transaction.uuid
+			});
+
+			req.done( function () {
```
westonruter: Should the response have the sanitized JS values to push back into the settings?

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -404,40 +628,30 @@ public function wp_redirect_status( $status ) {
...
-			if ( empty( $this->_post_values ) ) { // if not isset or of JSON error
-				$this->_post_values = false;
-			}
-		}
-		if ( empty( $this->_post_values ) ) {
-			return array();
-		} else {
-			return $this->_post_values;
-		}
+		_deprecated_function( __METHOD__, '0.4.2', 'WP_Customize_Manager::transaction::data()' );
```
westonruter: Actually, `4.2.0`

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -404,40 +628,30 @@ public function wp_redirect_status( $status ) {
...
 	 * @return string|mixed $post_value Sanitized value or the $default provided
 	 */
 	public function post_value( $setting, $default = null ) {
-		$post_values = $this->unsanitized_post_values();
-		if ( array_key_exists( $setting->id, $post_values ) ) {
-			return $setting->sanitize( $post_values[ $setting->id ] );
-		} else {
-			return $default;
-		}
+		_deprecated_function( __METHOD__, '0.4.2', 'WP_Customize_Manager::transaction::get()' );
```
westonruter: Actually, `4.2.0`

src/wp-includes/class-wp-customize-transaction.php
westonruter: Done in e1dd864

src/wp-includes/class-wp-customize-transaction.php
westonruter: Done in e9b1ab8

src/wp-admin/js/customize-controls.js
kienstra: This would be good for usability. The user will see exactly what the page looks like after the sanitization.

The implementation could be similar to customize-controls.js, line 78:
```
$.each( response.sanitized(), function( key, value ) { 
           api.settings.settings[ key ] = value;
});
```

westonruter: Or rather, they would see in the *controls* what is the actual value being used in the preview. I think it would involve:

```js
$.each( response.sanitized(), function( key, value ) { 
    var setting = api( key ), oldTransport = setting.transport;
    setting.transport = 'postMessage'; // so that we don't inadvertently cause infinite reload of preview, if JS sanitize != PHP sanitize
    setting.set( value );
    setting.transport = oldTransport;
});
```

src/wp-includes/class-wp-customize-transaction.php
```diff
@@ -0,0 +1,346 @@
...
+	 * @var WP_Customize_Manager
+	 */
+	public $manager;
+
+	/**
+	 * @since 4.2.0
+	 * @access public
+	 * @var string
+	 */
+	public $uuid;
```
kienstra: Possibly create a ```get_uuid()``` method for ```WP_Customize_Transaction```, instead of allowing public access to the ```uuid``` property.
It seems that you've considered this, and maybe you want public access.
Several files access the uuid property with ```transaction->uuid```:
```
wp-admin/customize.php
wp-includes/theme.php
wp-includes/class-wp-customize-manager.php
```

But a ```get_uuid()``` method might prevent bugs, and make it easier to change.

src/wp-includes/class-wp-customize-transaction.php
```diff
@@ -0,0 +1,346 @@
...
+		} else {
+			$setting_id = $setting->id;
+		}
+		/**
+		 * @var WP_Customize_Setting|null $setting
+		 * @var string $setting_id
+		 */
+
+		if ( ! isset( $this->data[ $setting_id ] ) ) {
+			// @todo Should this instead return $setting_obj->default? Or only if is_null( $default )?
```
kienstra: I think it would be better to return ```$setting_obj->default``` only if ```is_null( $default )```
It might be confusing to pass ```$default``` as an argument, but have a different default returned.

westonruter:
Closing in favor of #161.

PR #62, opened 2015-01-12
Trac 30988: Setting's default value doesn't apply on preview window

westonruter:
https://core.trac.wordpress.org/ticket/30988

Run unit tests via the new `customize` group:

```bash
phpunit --group customize
```


tests/phpunit/tests/customize/setting.php
```diff
@@ -0,0 +1,341 @@
...
+		$bar = new WP_Customize_Setting( $this->manager, 'bar', $args );
+		$this->assertEquals( 'bar', $bar->id );
+		foreach ( $args as $key => $value ) {
+			$this->assertEquals( $value, $bar->$key );
+		}
+		$this->assertEquals( 10, has_filter( "customize_sanitize_{$bar->id}", $args['sanitize_callback'] ) );
+		$this->assertEquals( 10, has_filter( "customize_sanitize_js_{$bar->id}" ), $args['sanitize_js_callback'] );
+	}
+
+	function test_preview_standard_types() {
```
westonruter: From @ocean90:

> Tests: It's not clear enough which test is specific for this issue. Also there are to many assertions in one tests. These should be splitted up. Closures should be removed too.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -399,23 +399,36 @@ public function wp_redirect_status( $status ) {
...
+	 * Decode the $_POST['customized'] values, and cache in _post_values array.
 	 *
-	 * @since 3.4.0
-	 *
-	 * @param WP_Customize_Setting $setting A WP_Customize_Setting derived object
-	 * @return string $post_value Sanitized value
+	 * @since 4.1.1
 	 */
-	public function post_value( $setting ) {
+	protected function parse_post_data() {
```
westonruter: From @ocean90:

> `parse_post_data()`: Can't we introduce `private function post_values()` and use `$this->post_values()`; instead of `$this->parse_post_data(); $this->_post_values;`?

tests/phpunit/tests/customize/setting.php
westonruter: @ocean90 I broke then up in half in 2c92f831d04626ef661b0b1a77a6885d286db0d3

src/wp-includes/class-wp-customize-manager.php
westonruter: @ocean90 I re-integrated these in 39d5d8f34dbee10c8883d9e1be3f84adab2987bc

tests/phpunit/tests/customize/setting.php
westonruter: Removed closures in these commits: https://github.com/xwp/wordpress-develop/compare/c4a0da93112e942ab35d6378c467e3d2fdde8ac5...edf7836af3bed880502b4ed5050fbb294be05fb4

westonruter: Also added specific test case for the issue at hand: https://github.com/xwp/wordpress-develop/commit/2da22194ecceb45eaf2a4b404beefb363c4fe40a

PR #67, opened 2015-02-01
Trac #30936: Dynamically create WP_Customize_Settings for settings created on JS client

westonruter:
https://core.trac.wordpress.org/ticket/30936

Builds on top of patch for Trac 30988 (Setting's default value doesn't apply on preview window). See: https://github.com/xwp/wordpress-develop/pull/62

Run unit tests via the new `customize` group:

```bash
phpunit --group customize
```


src/wp-includes/class-wp-customize-manager.php
```diff
@@ -111,10 +112,22 @@ public function __construct() {
 	 *
 	 * @since 3.4.0
 	 *
+	 * @param string|null $action whether the supplied Ajax action is being run (since 4.2.0).
```
westonruter: From @ocean90:

> the new param for doing_ajax should be added via the @since tag, you can remove the hint from @param afterwards.

westonruter: Done here: https://github.com/xwp/wordpress-develop/commit/95d1ef6ce856079491174959d35796ab806e6770

westonruter: Actually, here: 605569f

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -726,6 +739,62 @@ public function add_setting( $id, $args = array() ) {
...
+			 * Filter a dynamic setting's constructor args.
+			 *
+			 * For a dynamic setting to be registered, this filter must be employed
+			 * to override the default false value with an array of args to pass to
+			 * the WP_Customize_Setting constructor.
+			 *
+			 * @since 4.2.0
+			 *
+			 * @param false|array $setting_args  The arguments to the WP_Customize_Setting constructor.
+			 * @param string $setting_id  ID for dynamic setting, usually coming from $_POST['customized'].
```
ocean90: These should be aligned.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -726,6 +739,62 @@ public function add_setting( $id, $args = array() ) {
...
+
+			/**
+			 * Allow non-statically created settings to be constructed with custom WP_Customize_Setting subclass.
+			 *
+			 * @since 4.2.0
+			 *
+			 * @param string $setting_class  WP_Customize_Setting or a subclass.
+			 * @param string $setting_id  ID for dynamic setting, usually coming from $_POST['customized'].
+			 * @param string $setting_args  WP_Customize_Setting or a subclass.
+			 */
```
ocean90: These should be aligned.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -726,6 +740,62 @@ public function add_setting( $id, $args = array() ) {
 	}
 
 	/**
+	 * Register any dynamically-created settings, such as those from $_POST['customized'] that have no corresponding setting created.
+	 *
+	 * This is a mechanism to "wake up" settings that have been dynamically created
+	 * on the frontend and have been added to a transaction. When the transaction is
+	 * loaded, the dynamically-created settings then will get created and previewed
+	 * even though they are not directly created statically with code.
+	 *
```
westonruter: Missing `@since`

src/wp-includes/class-wp-customize-manager.php
westonruter: Done in fe0791a

src/wp-includes/class-wp-customize-manager.php
westonruter: Done in fe0791a

src/wp-includes/class-wp-customize-widgets.php
```diff
@@ -95,193 +94,92 @@ public function __construct( $manager ) {
...
-	 * @since 3.9.0
-	 * @access public
-	 *
-	 * @param array $sidebars_widgets Associative array of sidebars and their widgets.
-	 * @return array Filtered array of sidebars and their widgets.
+	 * @param false|array $args
+	 * @param string $setting_id
+	 * @return false|array
 	 */
-	public function prepreview_added_sidebars_widgets( $sidebars_widgets ) {
```
westonruter: To prevent fatal errors with existing plugins which try to reference this method and other deprecated methods in this class, we may want to instead leave this function here but add a `@deprecated` tag to the PhpDoc and then replace the body with just `_deprecated_function( __METHOD__, '4.2.0' );`

westonruter:
Merged in 2518bbbb1f7a6fcb3a4c7e461967099c7b740fac. Thanks @ocean90!

PR #68, opened 2015-02-02
Trac 31196: Loading indicators for the Customizer preview

westonruter:
https://core.trac.wordpress.org/ticket/31196


westonruter:
Now in Core: https://core.trac.wordpress.org/changeset/31697

PR #70, opened 2015-02-11
#31295: Programmatically changing sidebars_widgets Customizer settings causes unexpected widget control focus

westonruter:
https://core.trac.wordpress.org/ticket/31295


westonruter:
Merged in https://core.trac.wordpress.org/changeset/31428

PR #72, opened 2015-02-26
31465: Fix JS for sidebar_widgets customizer control to not return undefined items in getWidgetFormControls method

westonruter:
https://core.trac.wordpress.org/ticket/31465


westonruter:
Merged in 61dfcc3b4ce6979da89e826b811dc63dbb60568a.

PR #74, opened 2015-03-02
#31484: Widgets in Customizer during theme preview fail to preview widgets with prior sidebars_widgets theme mod and fail to apply upon activation

westonruter:
https://core.trac.wordpress.org/ticket/31484

### DIFF
--- a/src/wp-admin/js/customize-controls.js
+++ b/src/wp-admin/js/customize-controls.js
@@ -17,6 +17,7 @@
 
 			this.id = id;
 			this.transport = this.transport || 'refresh';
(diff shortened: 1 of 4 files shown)


src/wp-admin/js/customize-controls.js
```diff
@@ -17,6 +17,7 @@
 
 			this.id = id;
 			this.transport = this.transport || 'refresh';
+			this._dirty = options.dirty || false;
```
westonruter: We might as well promote this to a full-fledged public `dirty` property, and make it a `Value` so we can listen to changes.

PR #81, opened 2015-04-04
#31884: Update widget customizer spinner to use visibility instead of display

westonruter:
https://core.trac.wordpress.org/ticket/31884

See 69832941614b90b965f3f952b3344508eb1e8c40 and https://core.trac.wordpress.org/ticket/22839


westonruter:
Committed to Core.

PR #82, opened 2015-04-04
#31885: Fix widget form synchronization for select[multiple]

westonruter:
https://core.trac.wordpress.org/ticket/31885


westonruter:
Committed to Core: https://core.trac.wordpress.org/ticket/31885#comment:5

PR #85, opened 2015-05-23
Trac #32474: Facilitate widgets to be stored in posts instead of options

westonruter:
https://core.trac.wordpress.org/ticket/32474

Required changes to support Widget Posts: https://github.com/xwp/wp-customize-widgets-plus/pull/12


westonruter:
Committed to Core: https://core.trac.wordpress.org/changeset/32602

PR #86, opened 2015-05-25
Trac #32496: Fix widget-update Ajax call error response

westonruter:
https://core.trac.wordpress.org/ticket/32496


westonruter:
Merged in https://core.trac.wordpress.org/changeset/32601

PR #87, opened 2015-05-26
Trac #30737: JS templates for Customizer Panels and Sections

westonruter:
https://core.trac.wordpress.org/ticket/30737

Note that this conflicts with https://github.com/xwp/wordpress-develop/pull/88


westonruter:
This has been committed to Core.

PR #88, opened 2015-05-29
#31336: Customizer: separate navigation from options UI for better UX by removing accordion behavior

westonruter:
https://core.trac.wordpress.org/ticket/31336

Note that this conflicts with https://github.com/xwp/wordpress-develop/pull/87


westonruter:
This has been merged.

PR #91, opened 2015-05-31
Menu Customizer integration

westonruter:


src/wp-content/themes/twentyfifteen/sidebar.php
```diff
@@ -10,7 +10,7 @@
 if ( has_nav_menu( 'primary' ) || has_nav_menu( 'social' ) || is_active_sidebar( 'sidebar-1' )  ) : ?>
 	<div id="secondary" class="secondary">
 
-		<?php if ( has_nav_menu( 'primary' ) ) : ?>
+		<?php if ( has_nav_menu( 'primary' ) || is_customize_preview() ) : ?>
```
ocean90: Can we move `is_customize_preview()` into `has_nav_menu()`?

westonruter: @ocean90 ok, how about 01cd9a329b9debda057b545de3c77c5f3bc6f209?

westonruter:
> The `fallback_cb` change is interesting. Shouldn't we move this to wp_nav_menu()?

So something like modify the `$defaults`:

```php
$defaults = array( 
    /* ... */ 
    'fallback_cb' => is_customize_preview() ? '' : 'wp_page_menu' 
    /* ... */ 
);
```

westonruter:
@ocean90:

> The `fallback_cb` change is interesting. Shouldn't we move this to wp_nav_menu()?

So modify the `$defaults`?

```php
$defaults = array( 
    /* ... */ 
    'fallback_cb' => is_customize_preview() ? '' : 'wp_page_menu' 
    /* ... */ 
);
```

src/wp-includes/nav-menu.php
```diff
@@ -151,6 +156,10 @@ function has_nav_menu( $location ) {
 		return false;
 	}
 
+	if ( is_customize_preview() ) {
+		return true;
+	}
```
westonruter: I think actually this shouldn't be included at all. For Widget Customizer, the approach we took instead is adding a filter to `is_active_sidebar()`. We need a similar filter `has_nav_menu` so we can pick up whether or not a nav menu location is being referenced in the preview so that we can refresh the preview when a location is being updated, but no menu was previously assigned here. So this needs to be replaced with:

```php
$has_nav_menu = ( ! empty( $locations[ $location ] );
/**
 * @todo Docs.
 */
return apply_filters( 'wp_nav_menu', $has_nav_menu, $location );
```

westonruter: Done in c5fa537

src/wp-includes/nav-menu-template.php
```diff
@@ -230,6 +230,11 @@ function wp_nav_menu( $args = array() ) {
 	'echo' => true, 'fallback_cb' => 'wp_page_menu', 'before' => '', 'after' => '', 'link_before' => '', 'link_after' => '', 'items_wrap' => '<ul id="%1$s" class="%2$s">%3$s</ul>',
 	'depth' => 0, 'walker' => '', 'theme_location' => '' );
 
+	// Prevent a fallback_cb in Customizer Preview to assist with has_nav_menu() and partial refresh.
+	if ( is_customize_preview() ) {
+		$defaults['fallback_cb'] = '';
+	}
+
```
ocean90: This doesn't seem right because you just override the default value, which is already an empty string. This should probably be placed in line 250.
But then I noticed, that we have a `wp_nav_menu_args` filter here too. Can't we use the filter here too?

westonruter: @ocean90 What we want is to prevent `wp_page_menu` to not be used when someone just does:

```php
wp_nav_menu( array( 'theme_location' => 'primary' ) );
```

But if they *do* actually supply this fallback callback, then we should honor it:

```php
wp_nav_menu( array( 'theme_location' => 'primary', 'fallback_cb' => 'wp_page_menu' ) );
```

And so we can't use `wp_nav_menu_args` filter because we don't know if they supplied that explicitly, or if it was from the defaults.

PR #92, opened 2015-06-26
#30677: Update the preview URL when changing the related show_on_front setting

westonruter:
https://core.trac.wordpress.org/ticket/30677


westonruter:
Committed to Core in r32976.

PR #98, opened 2015-07-06
 #32814: Custom Menu Widget should show newly added menus in Customize

westonruter:
https://core.trac.wordpress.org/ticket/32814


src/wp-admin/js/customize-nav-menus.js
```diff
@@ -2370,6 +2399,32 @@
...
+					}
+					var select, oldMenuOption, oldMenuSelected, newMenuOption;
+					select          = setting.container.find( 'select' );
+					oldMenuOption   = select.find( 'option[value=' + String( update.previous_term_id ) + ']' );
+					oldMenuSelected = select.find( 'option[value=' + String( update.previous_term_id ) + ']:selected' );
+					newMenuOption   = select.find( 'option[value=' + String( update.term_id ) + ']' );
+
+					// Adjust menu options matching the old ID, setting them to the new ID.
+					if ( oldMenuSelected.length !== 0 ) {
+						newMenuOption.attr( 'selected', true );
```
westonruter: @adamsilverstein I think then you need to do `select.trigger( 'change' )`, which would then result in the widget setting getting updated. 

Or perhaps it would be better to handle this on the server… if we update the setting like this on the client, then the result is that the setting is going to become dirty here meaning the user would have to re-save. But if we handle it on the server during the `customize_save` action. When the menus get inserted, we need to update the Custom Menu widgets at that time. Then the `customize_save_response` filter can include a `data.custom_menu_widget_updates` just like we're including `data.nav_menu_updates`, and then for each `data.custom_menu_widget_updates` it can update the settings with those values, and then clear out the dirty state.

westonruter:
Merged in r33488.

PR #100, opened 2015-07-13
Trac #32769: Customizer Menus: Setting sanitization innefficiencies resulting in severe performance slowdowns

westonruter:
https://core.trac.wordpress.org/ticket/32769


westonruter:
Committed to Core in r33256.

PR #102, opened 2015-07-18
#32810: Customizer Menus: Indicate when there are no items

westonruter:
https://core.trac.wordpress.org/ticket/32810


src/wp-admin/js/customize-nav-menus.js
```diff
@@ -316,6 +316,13 @@
 				var items, typeInner;
 				items = data.items;
 				if ( 0 === items.length ) {
+					if ( 0 === self.pages[type] ) {
+						$( '#available-menu-items-' + type )
+							.addClass( 'cannot-expand' )
+							.removeClass( 'loading' )
+							.find( '.accordion-section-title > button' )
+							.prop( 'tabIndex', -1 );
```
westonruter: @valendesigns Should this `tabIndex` be reset to `0` if there are results?

valendesigns: How will it all of the sudden have results.

westonruter: Thanks for clarifying over Slack.

PR #118, opened 2015-09-06
Trac 33665 Menu Customizer: Implement indicators for invalid menu items

kucrut:
Work in progress.


kucrut:
@westonruter Could you please review this? I'm still not sure about the approach I took to expose the `_invalid` property. Also, I think it's best if we don't preview the changes made to invalid menu items, since they're not displayed on the front-end anyway. What do you think?

westonruter:
@kucrut sorry, the `master` branch wasn't up to date and so a lot of other commits got added here. I updated `master` but for some reason GitHub is not picking up on the change :-( Could you open a new PR please?

PR #119, opened 2015-09-06
Trac 33665: Menu Customizer: Implement indicators for invalid menu items 

kucrut:
Originally https://github.com/xwp/wordpress-develop/pull/118


kucrut:
@westonruter Here you go.

src/wp-includes/class-wp-customize-nav-menus.php
```diff
@@ -537,6 +537,11 @@ public function customize_register() {
 
 				$value = (array) $item;
 				$value['nav_menu_term_id'] = $menu_id;
+
+				if ( ! isset( $value['_invalid'] ) ) {
+					$value['_invalid'] = false;
+				}
```
westonruter: @kucrut won't this always be set by `wp_get_nav_menu_items` via `wp_setup_nav_menu_item`?

westonruter:
> Could you please review this? I'm still not sure about the approach I took to expose the `_invalid` property. 

I think you have the defaults getting set properly, but I think the actual value for `_invalid` needs to be set inside of `WP_Customize_Nav_Menu_Item_Setting::populate_value()`. In there I think it needs to do the logic found in `wp_setup_nav_menu_item()` for setting the `_invalid` property, if `$this->value['_invalid']` is not set.

> Also, I think it's best if we don't preview the changes made to invalid menu items, since they're not displayed on the front-end anyway. What do you think?

Yeah, I think you are right. If you look at `wp_get_nav_menu_items()` it actually removes invalid menu items if not in the admin anyway:

```php
	if ( ! is_admin() ) { // Remove invalid items only in frontend
		$items = array_filter( $items, '_is_valid_nav_menu_item' );
	}
```

src/wp-includes/class-wp-customize-setting.php
```diff
@@ -1039,6 +1039,8 @@ public function filter_wp_get_nav_menu_items( $items, $menu, $args ) {
 		$should_remove = (
 			false === $this_item
 			||
+			true === $this_item['_invalid']
+			||
```
westonruter: This _may_ be redundant given the existing logic in `wp_get_nav_menu_items()`:

```php
	if ( ! is_admin() ) { // Remove invalid items only in frontend
		$items = array_filter( $items, '_is_valid_nav_menu_item' );
	}
```

But it wouldn't hurt to be explicit here.

src/wp-includes/class-wp-customize-nav-menus.php
kucrut: No, valid items don't have `_invalid`.

kucrut:
> Also, I think it's best if we don't preview the changes made to invalid menu items, since they're not displayed on the front-end anyway. What do you think?

@westonruter Invalid items are already removed from the preview. What I really meant was to *not* make the state *dirty* when an invalid item changes.

kucrut:
> I think you have the defaults getting set properly, but I think the actual value for `_invalid` needs to be set inside of `WP_Customize_Nav_Menu_Item_Setting::populate_value()`. In there I think it needs to do the logic found in `wp_setup_nav_menu_item()` for setting the `_invalid` property, if `$this->value['_invalid']` is not set.

Setting `_invalid` in `WP_Customize_Nav_Menu_Item_Setting::populate_value()` would be too late and will cause an exception thrown by its `__construct()` because the initial `$value` passed don't always have `_invalid`.

westonruter:
@kucrut like for unit tests I think we should test that the changes you've introduced are having the desired impact.

westonruter:
Committed to Core: https://core.trac.wordpress.org/changeset/35302

Thanks, @kucrut!

PR #127, opened 2015-11-09
Customize: Fix autofocus on `nav_menu_item` controls.

westonruter:
Embed the `nav_menu_item` control not only when the contained `nav_menu` section expands, but also if the control was autofocused. Also applies change from [33596] to workaround a broken `:focusable` selector in jQuery UI.

See #33258.


westonruter:
Committed to `trunk` in https://github.com/xwp/wordpress-develop/commit/01b19a634502cce209b8ae569c40391c3727d952

PR #133, opened 2015-11-19
Trac 34738: Customizer preview elements lose or misplace their preview data

westonruter:
https://core.trac.wordpress.org/ticket/34738

- [x] Add unit tests.


src/wp-includes/class-wp-customize-setting.php
```diff
@@ -183,6 +192,9 @@ protected function aggregate_multidimensional() {
 			self::$aggregated_multidimensionals[ $this->type ] = array();
 		}
 		if ( ! isset( self::$aggregated_multidimensionals[ $this->type ][ $id_base ] ) ) {
+			// Only add one action per multidimensional value.
```
westonruter: Only add one action per multidimensional ID base.

src/wp-includes/class-wp-customize-setting.php
```diff
@@ -183,6 +192,9 @@ protected function aggregate_multidimensional() {
...
+			add_action( 'customize_post_value_set', array( $this, '_clear_aggregated_multidimensional_perview_applied_flag' ) );
```
westonruter: Spelling mistake.

westonruter: Should we just use `customize_post_value_set_{$this->id}` instead? That will result in fewer actions triggered, but more actions added.

marcperel:
I've run this a few times running Twenty16 and LayersWP and it checks out for me. Good work Weston.

westonruter:
@marcperel thank you very much for testing. This has been committed to `trunk`: https://core.trac.wordpress.org/changeset/35724

PR #137, opened 2015-12-29
Trac 32681

MattGeri:
Enable shift clicking on a menu which then opens the menu for editing in the Customizer

https://core.trac.wordpress.org/ticket/32681

Hey guys, this is my first core ticket and my first piece of work on the Customizer. I'm still learning how everything fits together with the Customizer so please point out anything I may have done wrong, misunderstood, or is blatantly silly :wink:

Appreciate any feedback!


westonruter:
@MattGeri the patch is looking good. However, instead of focusing on the nav menu control (and implementing `focusMenuControl`), I think actually it should be focusing on the nav menu _item_. For example, each nav menu item is (normally) represented like:

```html
<li id="menu-item-7295" class="menu-item menu-item-type-post_type menu-item-object-page menu-item-7295"><a href="https://example.com/about/">About</a></li>
```

The underlying nav menu item post ID (7295) is mentioned twice in the `li` element. So I suggest eliminating the `selector = menuSelectors.join( ',' );` you have and instead just use `.menu-item` as the delegated event selector, and then attempt to parse out the post ID from the element ID or from the class name. With this in hand, then you can do:

```js
api.preview.send( 'focus-nav-menu-item-control', navMenuItemId );
```

And this would then result in the existing `focusMenuItemControl` method being called, once you bind it to that message being sent from the preview.

By switching to focus on the nav menu items instead of the nav menus, I think the `highlightControls` would then get simplified quite a bit.

src/wp-includes/class-wp-customize-nav-menus.php
```diff
@@ -950,6 +950,9 @@ public function export_preview_data() {
 			),
 			'previewCustomizeNonce' => wp_create_nonce( 'preview-customize_' . $this->manager->get_stylesheet() ),
 			'navMenuInstanceArgs'   => $this->preview_nav_menu_instance_args,
+			'l10n'               => array(
+				'menuTooltip' => __( 'Shift-click to edit this menu.' ),
```
westonruter: I see this follows the pattern for widgets. The l10n string ID for widgets was not great. When switching to focus on nav menu items instead, I suggest `editNavMenuItemTooltip`.

MattGeri:
Hey @westonruter,

I've simplified the code and it's working quite nicely. Would love your feedback!

Thanks,
Matt

westonruter:
@MattGeri thanks. I reviewed the changes and tried them out. A couple key issues:

* Focusing on sub nav-menu items was not working.
* Focusing on a control would often happen too early, before the menu item form had expanded.

I pushed up a few revisions to fix the issues and a few other improvements: https://github.com/xwp/wordpress-develop/compare/d40d97c...8e82007

I attached created the patch file via `git diff master... > 32681.0.diff` and attached it to the Trac ticket.

I'll commit it later today unless you see any issue.

westonruter:
Now in `trunk`: https://core.trac.wordpress.org/changeset/36383

PR #138, opened 2016-01-11
Trac 31195: Add a user-friendly way to preview site responsiveness in the Customizer

westonruter:
https://core.trac.wordpress.org/ticket/31195

### DIFF
--- a/src/wp-admin/customize.php
+++ b/src/wp-admin/customize.php
@@ -137,7 +137,7 @@ do_action( 'customize_controls_print_scripts' );
 					<button class="customize-help-toggle dashicons dashicons-editor-help" aria-expanded="false"><span class="screen-reader-text"><?php _e( 'Help' ); ?></span></button>
 				</div>
 				<div class="customize-panel-description"><?php
-					_e( 'The Customizer allows you to preview changes to your site before publishing them. You can also navigate to different pages on your site to preview them.' );
+					_e( 'The Customizer allows you to preview changes to your site before publishing them. You can navigate to different pages on your site to preview them.' );
 				?></div>
 			</div>
 
@@ -148,6 +148,11 @@ do_action( 'customize_controls_print_scripts' );
 		</div>
 
 		<div id="customize-footer-actions" class="wp-full-overlay-footer">
+			<div class="devices">
+				<button type="button" class="preview-desktop active" aria-pressed="true" data-device="desktop"><span class="screen-reader-text"><?php _e( 'Enter desktop preview mode' ); ?></span></button>
+				<button type="button" class="preview-tablet" aria-pressed="false" data-device="tablet"><span class="screen-reader-text"><?php _e( 'Enter tablet preview mode' ); ?></span></button>
+				<button type="button" class="preview-mobile" aria-pressed="false" data-device="mobile"><span class="screen-reader-text"><?php _e( 'Enter mobile preview mode' ); ?></span></button>
+			</div>
 			<button type="button" class="collapse-sidebar button-secondary" aria-expanded="true" aria-label="<?php esc_attr_e( 'Collapse Sidebar' ); ?>">
 				<span class="collapse-sidebar-arrow"></span>
 				<span class="collapse-sidebar-label"><?php _e( 'Collapse' ); ?></span>
--- a/src/wp-admin/js/customize-controls.js
+++ b/src/wp-admin/js/customize-controls.js
@@ -3603,18 +3603,18 @@
 		});
 
 		// Preview size toggles.
-		$( '#customize-footer-actions .devices button' ).on( 'click keydown', function( event ) {
-			if ( api.utils.isKeydownButNotEnterEvent( event ) ) {
-				return;
-			}
-
-			$( '#customize-footer-actions .devices button' ).removeClass( 'active' );
+		$( '#customize-footer-actions .devices button' ).on( 'click', function( event ) {
+			var overlay = $( '.wp-full-overlay' ),
+			    device = $( event.currentTarget ).data( 'device' );
+			$( '#customize-footer-actions .devices button' ).removeClass( 'active' )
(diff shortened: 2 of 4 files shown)


src/wp-admin/js/customize-controls.js
```diff
@@ -3602,6 +3602,19 @@
...
+		$( '#customize-footer-actions .devices button' ).on( 'click', function( event ) {
+			var overlay = $( '.wp-full-overlay' ),
+			    device = $( event.currentTarget ).data( 'device' );
+			$( '#customize-footer-actions .devices button' ).removeClass( 'active' )
+			                                                .attr( 'aria-pressed', false );
+			overlay.removeClass( 'preview-full preview-desktop preview-tablet preview-mobile' )
+			       .toggleClass( 'preview-' + device );
+			$( event.currentTarget ).addClass( 'active' )
+			                        .attr( 'aria-pressed', true );
+			api.previewedDevice = device;
```
westonruter: `api.previewedDevice` should be an instance of `wp.customize.Value` so and so this would be changed to:

```js
wp.customize.previewedDevice.set( device );
```

The preceding code can then be moved into a handler such as this:

```
api.previewedDevice.bind( function( newDevice ) {
	var overlay = $( '.wp-full-overlay' );
	$( '#customize-footer-actions .devices button' ).removeClass( 'active' ).attr( 'aria-pressed', false );
	overlay.removeClass( 'preview-full preview-desktop preview-tablet preview-mobile' ).addClass( 'preview-' + newDevice );
	$( '#customize-footer-actions .devices' ).find( '.button-' + newDevice ).addClass( 'active' ).attr( 'aria-pressed', true );
} );
```

src/wp-admin/customize.php
```diff
@@ -148,6 +148,11 @@
 		</div>
 
 		<div id="customize-footer-actions" class="wp-full-overlay-footer">
+			<div class="devices">
+				<button type="button" class="preview-desktop active" aria-pressed="true" data-device="desktop"><span class="screen-reader-text"><?php _e( 'Enter desktop preview mode' ); ?></span></button>
+				<button type="button" class="preview-tablet" aria-pressed="false" data-device="tablet"><span class="screen-reader-text"><?php _e( 'Enter tablet preview mode' ); ?></span></button>
+				<button type="button" class="preview-mobile" aria-pressed="false" data-device="mobile"><span class="screen-reader-text"><?php _e( 'Enter mobile preview mode' ); ?></span></button>
+			</div>
```
westonruter: I suggest we add as `WP_Customize_Manager::get_previewed_devices()` which returns an array:

```php
function get_previewed_devices() {
	$devices = array(
		'desktop' => array(
			'label' => __( 'Enter desktop preview mode' ),
			'default' => true,
		),
		'tablet' => array(
			'label' => __( 'Enter tablet preview mode' ),
		),
		'mobile' => array(
			'label' => __( 'Enter mobile preview mode' ),
		),
	);
	$devices = apply_filters( 'customize_previewed_devices', $devices );
	return $devices;
}
```

And that this array be iterated over to build up this list, and maybe also allow plugins to extend which the devices that are previewable for the theme.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -1701,6 +1701,7 @@ public function customize_pane_settings() {
 			'autofocus' => array(),
 			'documentTitleTmpl' => $this->get_document_title_template(),
 		);
+		$this->add_setting( 'previewedDevice', array( 'default' => 'desktop' ) );
```
westonruter: This should be removed. It's not actually being used (I don't see `wp.customize( 'previewedDevice' )`), amd the previewed device isn't a _Customizer_ setting, but a UI configuration. It could be a _user_ setting, e.g. via `get_user_setting()`/`set_user_setting()`, to persist the user's previewed device.

Also, perhaps the default should be `mobile` if `wp_is_mobile()` returns true.

src/wp-admin/js/customize-controls.js
```diff
@@ -3602,6 +3602,19 @@
 			event.preventDefault();
 		});
 
+		// Preview size toggles.
+		$( '#customize-footer-actions .devices button' ).on( 'click', function( event ) {
+			var overlay = $( '.wp-full-overlay' ),
+			    device = $( event.currentTarget ).data( 'device' );
+			$( '#customize-footer-actions .devices button' ).removeClass( 'active' )
+			                                                .attr( 'aria-pressed', false );
+			overlay.removeClass( 'preview-full preview-desktop preview-tablet preview-mobile' )
```
westonruter: The list of device class names to remove could be derived from `WP_Customize_Manager::get_previewed_devices()` if exported to JS in `WP_Customize_Manager::customize_pane_settings()`.

src/wp-admin/customize.php
celloexpressions: Good call. I'm implementing this as `get_previewable_devices()` to mitigate potential confusion with the device with which the customizer is open. Developers can also now turn the feature off entirely with:
```
add_filter( 'customize_previewable_devices', function() { return array(); } );
```

src/wp-includes/class-wp-customize-manager.php
celloexpressions: I believe `wp_is_mobile()` is really more about whether the device supports touch and can't be easily relied upon, so thinking it's probably better to default to desktop for now since the feature is only available on larger screens (for now). A persistent user setting probably wouldn't give much benefit since I'm guessing most users would only switch into/out of these modes when they're intending to look at a particular thing, then want to switch back. I could see some users getting "stuck" in mobile preview and becoming confused.

src/wp-admin/js/customize-controls.js
celloexpressions: That's much better, about to post the revised patch.

PR #140, opened 2016-01-23
Trac 18584: Nav menus need more hooks for extensibility (on admin page & in customizer)

westonruter:
https://core.trac.wordpress.org/ticket/18584

This extends https://github.com/xwp/wordpress-develop/pull/139

### DIFF
--- a/src/wp-includes/customize/class-wp-customize-nav-menu-item-control.php
+++ b/src/wp-includes/customize/class-wp-customize-nav-menu-item-control.php
@@ -85,44 +85,58 @@ class WP_Customize_Nav_Menu_Item_Control extends WP_Customize_Control {
 		</div>
 
 		<div class="menu-item-settings" id="menu-item-settings-{{ data.menu_item_id }}">
+			<?php $nav_menu_item_fields = array(); ?>
+			<?php ob_start(); ?>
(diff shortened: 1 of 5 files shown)


src/wp-includes/customize/class-wp-customize-nav-menu-item-control.php
```diff
@@ -85,51 +85,86 @@ public function content_template() {
...
+			 * @since 4.5.0
+			 *
+			 * @param array $nav_menu_item_fields Mapping of ID to the field paragraph HTML.
+			 * @param array $context {
+			 *     Context for applied filter.
+			 *
+			 *     @type WP_Customize_Nav_Menu_Item_Control $customize_control Customize control.
+			 * }
+			 */
+			$nav_menu_item_fields = apply_filters( 'wp_nav_menu_item_fields', $nav_menu_item_fields, array( 'customize_control' => $this ) );
```
celloexpressions: The customizer filter should use a different string than the admin. Maybe `wp_customize_nav_menu_item_fields`?

Additionally, in the filter docs here we should make it abundantly clear that this is a JS template and needs to be handled differently from the PHP version for the admin page because menu item controls are dynamically added.

PR #141, opened 2016-02-15
trac-34755 : Change string in '(Currently set to: %s)' to location name

kienstra:
This PR for [trac-34755](https://core.trac.wordpress.org/ticket/34755) changes the "Menu Location" string in Customizer menu controls to the location name. Before, this string was the location slug.

So this adds an object to the data passed to the Customizer : `locationSlugMappedToName`. And uses this to retrieve the location name.

This also changes the argument name from `themeLocations` to `themeLocationSlugs`. This doesn't change the argument, it only clarifies the fact that it consists of slugs.

### DIFF
--- a/src/wp-admin/js/customize-nav-menus.js
+++ b/src/wp-admin/js/customize-nav-menus.js
@@ -805,19 +805,20 @@
 		/**
 		 * @param {array} themeLocations
 		 */
-		updateAssignedLocationsInSectionTitle: function( themeLocations ) {
+		updateAssignedLocationsInSectionTitle: function( themeLocationSlugs ) {
 			var section = this,
 				$title;
 
 			$title = section.container.find( '.accordion-section-title:first' );
 			$title.find( '.menu-in-location' ).remove();
-			_.each( themeLocations, function( themeLocation ) {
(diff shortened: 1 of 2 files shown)


src/wp-admin/js/customize-nav-menus.js
```diff
@@ -805,19 +805,20 @@
...
 			var section = this,
 				$title;
 
 			$title = section.container.find( '.accordion-section-title:first' );
 			$title.find( '.menu-in-location' ).remove();
-			_.each( themeLocations, function( themeLocation ) {
+			_.each( themeLocationSlugs, function( themeLocationSlug ) {
 				var $label = $( '<span class="menu-in-location"></span>' );
-				$label.text( api.Menus.data.l10n.menuLocation.replace( '%s', themeLocation ) );
+				var locationName = api.Menus.data.locationSlugMappedToName[ themeLocationSlug ];
```
westonruter: Let's combine the two `var` declarations, even just putting at the top of the function: `var $label, locationName;`

Also, at the top of the file, where `api.Menus.data` is defined, add `locationSlugMappedToName` with an initial value of `{}`.

PR #143, opened 2016-02-24
trac-34747 To Master : Add Filter For Customizer Description

kienstra:
This pull request for [trac-34747](https://core.trac.wordpress.org/ticket/34747) adds a filter for the Customizer description. This appears at the top of the Customizer's primary panel, after the title.

If there's no description, it doesn't echo the `?` button, which displays it.

Please don't merge this PR, as the "master" branch needs to be in sync with the remote [WordPress repo](https://core.trac.wordpress.org/browser/trunk).

### DIFF
--- a/src/wp-admin/customize.php
+++ b/src/wp-admin/customize.php
@@ -133,11 +133,27 @@ do_action( 'customize_controls_print_scripts' );
 				<div class="accordion-section-title">
 					<span class="preview-notice"><?php
 						echo sprintf( __( 'You are customizing %s' ), '<strong class="panel-title site-title">' . __( 'Loading&hellip;' ) . '</strong>' );
-					?></span>
-					<button class="customize-help-toggle dashicons dashicons-editor-help" aria-expanded="false"><span class="screen-reader-text"><?php _e( 'Help' ); ?></span></button>
+					?></span><?php
+
+					/**
+					 * Filter the description in the primary Customizer panel.
+					 *
+					 * @since 4.6
+					 *
+					 * @param string $customizer_description Appears at top of primary Customizer panel, under Customizer title.
+					 */
+					$customizer_description = apply_filters( 'customizer_primary_panel_description', __( 'The Customizer allows you to preview changes to your site before publishing them. You can also navigate to different pages on your site to preview them.' ) );
+					if ( ! empty( $customizer_description ) ) {
+						?>
+						<button class="customize-help-toggle dashicons dashicons-editor-help" aria-expanded="false">
+							<span class="screen-reader-text"><?php _e( 'Help' ); ?></span>
+						</button>
+						<?php
+					}
+					?>
 				</div>
 				<div class="customize-panel-description"><?php
-					_e( 'The Customizer allows you to preview changes to your site before publishing them. You can also navigate to different pages on your site to preview them.' );
+					echo esc_html( $customizer_description );
 				?></div>
 			</div>
 


kienstra:
Sorry, I'm resolving the merge conflicts now.

westonruter:
I forgive you. :smile:

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -184,6 +184,33 @@
...
+	protected $original_customizer_title;
+
+	/**
+	 * Filtered title of the Customizer, shown in the top of the root panel.
+	 *
+	 * @since 4.6
+	 * @access protected
+	 * @var string
+	 */
+	protected $filtered_root_panel_title;
```
westonruter: @kienstra I think this should instead be `$root_panel_title`, and its initial value can be set at the top in `wp-admin/customize.php` to be `get_bloginfo( 'name' )`.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -184,6 +184,33 @@
 	protected $return_url;
 
 	/**
+	 * Original title of the Customizer, before being filtered.
+	 *
+	 * @since 4.6
+	 * @access protected
+	 * @var string
+	 */
+	protected $original_customizer_title;
```
westonruter: I think this can be removed, just letting `$root_panel_title` be `null` initially, and defined explicitly in the top of `wp-admin/customize.php` as is done for setting the `$preview_url`, `$return_url`, and `$autofocus`.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -184,6 +184,33 @@
...
+	protected $filtered_root_panel_title;
+
+	/**
+	 * Filtered description of the Customizer, below the title.
+	 *
+	 * @since 4.6
+	 * @access protected
+	 * @var string
+	 */
+	protected $filtered_root_panel_description;
```
westonruter: Per above, let's have this just be `$root_panel_description`

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -1601,6 +1628,93 @@ public function get_return_url() {
...
 
 	/**
+	 * Set Customizer title at top of root Customizer panel.
+	 *
+	 * Appears after text "You are customizing."
+	 *
+	 * @since 4.6
+	 * @access public
+	 */
+	public function set_filtered_root_panel_title() {
```
westonruter: Rename this to `set_root_panel_title()`

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -1601,6 +1628,93 @@ public function get_return_url() {
...
+		$this->original_customizer_title = get_bloginfo( 'name' );
```
westonruter: This can be removed.

src/wp-includes/class-wp-customize-manager.php
westonruter: Also, it should accept a `$root_panel_title` as its argument.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -1601,6 +1628,93 @@ public function get_return_url() {
...
+
+		/**
+		 * Filter title in root Customizer panel.
+		 *
+		 * @since 4.6
+		 *
+		 * @param string $customizer_title Appears at top of root Customizer panel.
+		 */
+		$this->filtered_root_panel_title = apply_filters( 'customize_root_panel_title', $this->original_customizer_title );
```
westonruter: Let's filter in the getter, not in the setter. So this method can be merely:

```php
$this->root_panel_title = $root_panel_title;
```

Where `$root_panel_title` is passed in as a method argument.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -1601,6 +1628,93 @@ public function get_return_url() {
...
+
+	/**
+	 * Get Customizer title at top of root Customizer panel
+	 *
+	 * Appears above the description.
+	 *
+	 * @since 4.6
+	 * @access public
+	 */
+	public function get_filtered_root_panel_title() {
```
westonruter: Rename to `get_root_panel_title()`.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -1601,6 +1628,93 @@ public function get_return_url() {
...
+		if ( isset( $this->filtered_root_panel_title ) ) {
+			return $this->filtered_root_panel_title;
+		}
```
westonruter: Replace this conditional with (along with the necessary phpdoc):

```php
return apply_filters( 'customize_root_panel_title', $this->root_panel_title );
```

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -1601,6 +1628,93 @@ public function get_return_url() {
...
+	 * Whether the Customizer root panel title has been filtered.
+	 *
+	 * True if this title is different from the original title.
+	 *
+	 * @since 4.6
+	 * @access public
+	 *
+	 * @return bool
+	 */
+	public function is_root_panel_title_filtered() {
```
westonruter: I see now why you have `original_customizer_title` so that you can pass back whether the title was filtered, to short-circuit the `blogname` updates also updating the root panel title.

Hummmm…

src/wp-includes/class-wp-customize-manager.php
westonruter: On second thought, how about this…

Let's have `$root_panel_title` remain `null` and not get set explicitly…

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -1727,6 +1845,9 @@ public function customize_pane_settings() {
 			'documentTitleTmpl' => $this->get_document_title_template(),
 			'previewableDevices' => $this->get_previewable_devices(),
 			'selectiveRefreshEnabled' => isset( $this->selective_refresh ),
+			'filteredRootPanelTitle' => $this->get_filtered_root_panel_title(),
+			'isRootPanelTitleFiltered' => $this->is_root_panel_title_filtered(),
```
westonruter: How about replacing this with:

```
`isRootPanelTitleOverridden` => ! is_null( $this->root_panel_title ),
```

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -1727,6 +1845,9 @@ public function customize_pane_settings() {
 			'documentTitleTmpl' => $this->get_document_title_template(),
 			'previewableDevices' => $this->get_previewable_devices(),
 			'selectiveRefreshEnabled' => isset( $this->selective_refresh ),
+			'filteredRootPanelTitle' => $this->get_filtered_root_panel_title(),
```
westonruter: We could leave `$this->root_panel_title` as `null`, but the `get_root_panel_title()` method can check if `$this->root_panel_title` is `null`, and if so return `get_bloginfo( 'name' )`.

src/wp-includes/class-wp-customize-manager.php
westonruter: Actually, how about this instead:

```php
return is_null( $this->root_panel_title ) ? get_bloginfo( 'name' ) : $this->root_panel_title;
```

So then I think we can actually eliminate the `customize_root_panel_title` filter entirely. Plugins can just call the method if they want to override the title instead of using a filter.

src/wp-includes/class-wp-customize-manager.php
westonruter: Yeah, per my other comments, I think we can use `isset( $this->root_panel_title )` to determine if a custom title has been supplied. If not, the getter will return the `get_bloginfo( 'name' )`

kienstra:
Hi @westonruter,
Thanks a lot for reviewing this PR. The latest commits apply your suggestions.

Though you didn't suggest changing `set_panel_description`, I [removed the filter](https://github.com/xwp/wordpress-develop/commit/5872ca49da2a4288ebf115cc0a39b8136138b20d#diff-4fb8a477f559bdfad2c1e6db6d1c8b06L1653) in this. This can be changed by passing a new description as an argument to [$wp_customize->set_panel_description](https://github.com/xwp/wordpress-develop/pull/143/files#diff-4fb8a477f559bdfad2c1e6db6d1c8b06R1655).

This is like you suggested for the [set_panel_title](https://github.com/xwp/wordpress-develop/commit/ded3b2c0a4b4e39466d38992b909ca01a459f400) method.

westonruter:
@kienstra I was hoping you would infer that. Good on ya!

kienstra:
Thanks, @westonruter :smiley:

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -1601,6 +1619,57 @@ public function get_return_url() {
...
+		$this->root_panel_title = $root_panel_title;
+	}
+
+	/**
+	 * Get Customizer title at top of root Customizer panel
+	 *
+	 * Appears above the description.
+	 *
+	 * @since 4.6
+	 * @access public
```
westonruter: Trivial thing: needs a `@return`.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -1601,6 +1619,57 @@ public function get_return_url() {
...
+		$this->root_panel_description = $root_panel_description;
+	}
+
+	/**
+	 * Get Customizer description, to be shown in the root panel.
+	 *
+	 * Appears after "You are customizing" and the Customizer title.
+	 *
+	 * @since 4.6
+	 * @access public
```
westonruter: Trivial thing: needs a `@return`.

src/wp-admin/customize.php
```diff
@@ -136,9 +136,9 @@
 					?></span>
 					<button class="customize-help-toggle dashicons dashicons-editor-help" aria-expanded="false"><span class="screen-reader-text"><?php _e( 'Help' ); ?></span></button>
 				</div>
-				<div class="customize-panel-description"><?php
-					_e( 'The Customizer allows you to preview changes to your site before publishing them. You can also navigate to different pages on your site to preview them.' );
-				?></div>
+				<div class="customize-panel-description">
+					<?php // Description produced with JavaScript ?>
+				</div>
```
westonruter: Curious as to the reason for using JS to populate this. Since it won't be later updated, should it just `echo $wp_customize->get_root_panel_description()`?

westonruter: Likewise, on line 137 above, this should be changed as:

```diff
- echo sprintf( __( 'You are customizing %s' ), '<strong class="panel-title site-title">' . get_bloginfo( 'name' ) . '</strong>' );
+ echo sprintf( __( 'You are customizing %s' ), '<strong class="panel-title site-title">' . $wp_customize->get_root_panel_title() . '</strong>' ); 
```

kienstra:
Hi @westonruter,
Thanks for reviewing these commits. As you suggested, I moved the output of the description to [customize.php](https://github.com/xwp/wordpress-develop/commit/d7c730170770bcea99221d5679fbd12015569214#diff-a325cc1f63ecc04fdb078079c6ec585bR145). And I [removed](https://github.com/xwp/wordpress-develop/commit/d7c730170770bcea99221d5679fbd12015569214#diff-58dfe22165ffd9650354f7bdd51002ecL3253) the JS that added this. 

Likewise, I moved the logic that hides the `.customize-help-toggle` button from [customize-controls.js](https://github.com/xwp/wordpress-develop/commit/d7c730170770bcea99221d5679fbd12015569214#diff-58dfe22165ffd9650354f7bdd51002ecL3257) to [customize.php](https://github.com/xwp/wordpress-develop/commit/d7c730170770bcea99221d5679fbd12015569214#diff-a325cc1f63ecc04fdb078079c6ec585bR138). This is determined by the [presence of the description](https://github.com/xwp/wordpress-develop/commit/d7c730170770bcea99221d5679fbd12015569214#diff-a325cc1f63ecc04fdb078079c6ec585bR139), so it should be in the same file that produces the description.

westonruter:
Great. Let's re-connect on this once 4.6 has kicked off.

kienstra:
Thanks, @westonruter! That sounds good.

PR #145, opened 2016-03-04
Trac 35947 : Customizer panel fails to fully expand leaving extra margin 

kienstra:
@westonruter,
This pull request for [trac-35497](https://core.trac.wordpress.org/ticket/35947) removes extra space at bottom of Customizer panel. As the [ticket](https://core.trac.wordpress.org/ticket/35947) notes, there's "unnecessary spacing when you go inside a panel." 

This occurs when there are many panels, using the code provided in the [ticket](https://core.trac.wordpress.org/ticket/35947) to produce the panels. The non-active sections and panels still occupied space in the DOM. So the active panel had a `margin-top` property to move it to the top.

This PR [sets](https://github.com/xwp/wordpress-develop/compare/master...trac-35947#diff-6bdd5041777c50e624f1d0beae2e0b7cR437) the height of the non-active sections to `0`. And it [doesn't add](https://github.com/xwp/wordpress-develop/compare/master...trac-35947#diff-58dfe22165ffd9650354f7bdd51002ecL1366) a `margin-top` value to the active panel's content.

I tested this with the test code on the [ticket](https://core.trac.wordpress.org/ticket/35947) on the following themes. There was no longer a space at the bottom of the panels. I couldn't find any side-effects. 
[Graphene](https://wordpress.org/themes/graphene/)
[Heuman](https://wordpress.org/themes/hueman/)
[Customizr](https://wordpress.org/themes/customizr)
[Twenty Fifteen](https://wordpress.org/themes/twentyfifteen)
[Twenty Sixteen](https://wordpress.org/themes/twentysixteen)

### DIFF
--- a/src/wp-admin/css/customize-controls.css
+++ b/src/wp-admin/css/customize-controls.css
@@ -430,6 +430,14 @@ h3.customize-section-title {
 	width: 300px;
 }
 
+.in-sub-panel #customize-info {
+	height: 0;
+}
+
+.in-sub-panel #customize-theme-controls > ul > .control-section:not(.current-panel) {
+	height: 0;
+}
+
 .in-sub-panel #customize-theme-controls .accordion-section.current-panel {
 	width: 100%;
 }
--- a/src/wp-admin/js/customize-controls.js
+++ b/src/wp-admin/js/customize-controls.js
@@ -1363,7 +1363,6 @@
 					content.parent().show();
 					position = content.offset().top;
 					scroll = container.scrollTop();
-					content.css( 'margin-top', ( headerActionsHeight - position - scroll ) );
 					accordionSection.addClass( 'current-panel' );
 					overlay.addClass( 'in-sub-panel' );
 					container.scrollTop( 0 );


westonruter:
Woah. Hey, @delawski, thoughts?

delawski:
Hey, @westonruter. I took a look at the code. I'm a bit concerned about removing the code responsible for repositioning panels with `margin-top`. It might not be very safe in current panel/section implementation, so I have to check it locally. I'll checkout the branch tomorrow and do few tests. I'll be able to tell something more then.

delawski:
**Different approach proposal**

I have pushed my proposal for the fix to the issue. Instead of removing `_recalculateTopMargin()` which is risky in current sliding panels implementation and [may cause new issues](https://core.trac.wordpress.org/ticket/35947#comment:6), I have delayed focus event on back button.

While working on [#34391](https://core.trac.wordpress.org/ticket/34391), I've noticed that focusing on element during CSS transition may cause unexpected results, very similar to the one presented. In case of [#34391](https://core.trac.wordpress.org/ticket/34391) I have implemented full solution taking advantage of `transitionEnd` event listening. In case of this ticket, I have used simpler solution - `focus()` is delayed by 180 ms which is the time needed for the transition to end.

I've tested the fix on Chrome/OS X, Firefox/OS X, Firefox/Windows 8.1, Firefox/Windows 7, IE11/Windows 7 (Twenty Sixteen in each case).

@kienstra, @westonruter: it'd be great if you could double check the fix works for you as expected. _As a side note: I was able to reproduce the original issue only in Firefox._

delawski:
I have created new branch and PR #146 so that the patch file doesn't contain previous commits and their reverts.

westonruter:
@delawski It is fine, even preferred, to have a PR include all of the activity and all commits here because the PR isn't actually merged. Having all of the commits shows all of the authors and makes it easier to give props. So I closed #146 in favor of keeping this PR.

kienstra:
**Tested Piotr's Commit**
_Works As Expected_

As [noted](https://core.trac.wordpress.org/ticket/35947#trac-change-14-1457327040610344) on [trac-35947](https://core.trac.wordpress.org/ticket/35947#trac-change-10-1457231855242489), @delawski's latest [commit](https://github.com/xwp/wordpress-develop/commit/89591f9e3c3ae42b6b1c40b350f30c51c90892fa) fixes the vertical spacing issue ([screenshot](https://drive.google.com/file/d/0B-0jvLc5J3NnWWFBN1BYa19FNEU/view?usp=sharing)). This issue was originally shown in [these](https://core.trac.wordpress.org/attachment/ticket/35947/inside-panel.png) [screenshots](https://core.trac.wordpress.org/attachment/ticket/35947/image-empty-space.png).

delawski:
I have pushed a fix proposal for the last issue. Please, take a look at my comment on trac.wordpress.org:

https://core.trac.wordpress.org/ticket/35947#comment:28

PR #149, opened 2016-04-04
Trac 36389: Fix selective refresh for widgets gets lost once `refresh` is used

westonruter:


westonruter:
This was merged in core.

PR #154, opened 2016-09-08
WIP - Trac 35395 (Custom CSS)

johnregan3:
**Opening PR for Feedback.  DO NOT MERGE.**

[35395](https://core.trac.wordpress.org/ticket/35395)

At this point, everything works (although there is a bug where changes display, then disappear in the Preview).  There is still plenty to do, but the implementation is functional.

Seeking feedback on the following questions in particular:

-  Class `WP_Custom_CSS`
     -  I used a Class for the following reasons:
          - It uses several methods specific to this feature (e.g., the transient)
          - It can be extended by theme/plugin authors.  
          - We can use class constants to make modifying the CPT slug and transient name simpler (DRY).  
     -  Where should this class be fired?  Currently, it's at the bottom of the file.
     -  Is it appropriate to use the CPT slug and transient name as constants/properties?
     -  Is the slug name "style" appropriate for this CPT?  I kept it separate from "custom-css" because it is quite possible that plugins are using this term.
     - I'd like to do CSSTidy and Custom CSS-specific Notifications within this class. Your thoughts?
-  JavaScript
     - Is the new code in `customize-controls.js` implemented correctly? (note that I still have a bug in the Preview to fix there)
- File placement
     - CodeMirror is found in `wp-includes/js/codemirror/`.  Is this appropriate?
     - Where should CSSTidy be located?

### Done
In this branch, I've done the following so far (with some foundational work from Nick Halsey):
- [x]  Created the section for Custom CSS in the customizer
- [x]  Created a new Controller, "Code Editor"
- [x]  Added foundational JS for the Code Editor control
- [x]  Installed CodeMirror
- [x]  Created a CPT called "style" to hold our custom CSS
     - Note: You can currently view the content of the posts at `wp-admin/edit.php?post_type=style`
- [x]  Created a custom Setting called "WP_Customize_CSS_Setting"
- [x]  Configured Setting to save to style CPT

### To Do
- [x]  Fix bug when updating the Preview (changes flash correctly for a moment, then disappear)
- [ ]  Implement CSSTidy
- [ ]  Implement Notifications
- [ ]  Add hooks
- [ ]  Escaping/Validation/Sanitization throughout
- [ ]  Make CPT hidden
- [ ] Test on Networks
- [ ]  Unit Tests

@westonruter @MattGeri @valendesigns @kienstra @delawski


src/wp-includes/class-wp-custom-css.php
```diff
@@ -0,0 +1,274 @@
...
+	public function register_post_type() {
+		register_post_type( self::POST_TYPE_SLUG, array(
+			'labels' => array(
+				'name' => __( 'Styles' ),
+				'singular_name' => __( 'Style' ),
+				'edit_item' => __( 'Edit Style' ),
+				'add_new_item' => __( 'Add New Style' ),
+			),
+			'public'  => true,
+			'show_ui' => true,
```
celloexpressions: This should probably be false. We'll add UI for revisions later, likely as part of snapshots/transactions and other customizer revisions.

src/wp-includes/class-wp-custom-css.php
```diff
@@ -0,0 +1,274 @@
...
+			'rewrite' => false,
+			'menu_position' => 50,
+			'query_var' => false,
+			'delete_with_user' => true,
+			'_builtin'    => true, /* internal use only. don't use this when registering your own post type. */
+
+			// @todo only "title" and "revisions" after testing.
+			'supports' => array( 'title', 'editor', 'author', 'revisions' ),
+			'capabilities' => array(
+				'delete_posts'           => 'edit_theme_options',
```
celloexpressions: We might consider using `edit_files` for non-multisite. Something to discuss in the core dev chat probably.

src/wp-includes/class-wp-custom-css.php
johnregan3: @celloexpressions Most definitely.  I just have it this way currently so I can view the post content when the post is saved. 😄

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -2307,6 +2310,50 @@ public function register_controls() {
...
+				__( 'link opens in a new window' )
+			),
+		) ) );
+
+		if ( ! is_multisite() ) {
+			$this->add_setting( 'wp_custom_css_more', array() );
+			$this->add_control( 'wp_custom_css_more', array(
+				'type'        => 'none',
+				'section'     => 'custom_css',
+				'description' => sprintf( '%s<br /><a href="%s" class="external-link" target="_blank">%s&nbsp;<span class="screen-reader-text">(%s)</span></a>.',
```
celloexpressions: I think we should probably make the parenthesis translatable as well.

src/wp-includes/customize/class-wp-customize-custom-css-setting.php
```diff
@@ -0,0 +1,125 @@
...
+/**
+ * Customize API: WP_Customize_Custom_CSS_Setting class
+ *
+ * @package WordPress
+ * @subpackage Customize
+ * @since 4.7.0
+ */
+
+/**
+ * A setting that is used to filter a value, but will not save the results.
```
celloexpressions: Need to update this description, although I'm also wondering whether this custom setting could be useful in plugins if it were a bit more generic. Something to think about.

celloexpressions:
This is a great start! I think `style` could run into conflicts as a post type slug as well. I wouldn't be opposed to doing `wp_custom_css` if we're concerned about plugin compatibility with `custom_css`. Plugins will need to update to provide a migration path to the core feature (and in cases where they want to, extend the core feature), so we may be able to get away with `custom_css` and some sort of a check to avoid overwriting existing styles if they follow the same post-name conventions.

I made a couple little tweaks and added some comments inline as well. The changes to `customize-controls.js` generally look correct, although the section-expansion binding could probably be improved. That should be the right place for CodeMirror, and CSSTidy would probably be in `wp-includes/CSSTidy`, unless we can get it down to one file, which could probably go in `wp-admin/includes/`, although I think that's less common and this seems like more of a `wp-includes` thing.

johnregan3:
@celloexpressions Most recent changes:

- Removed CSSTidy and CodeMirror (and all associated scripts, styles and JS control support)
- Mapped "unfiltered_css" to "unfiltered_html"
- Moved Notifications to the Setting class
- Various code cleanup/refactoring tasks
- Basic CSS Sanitization using wp_kses()

@westonruter you may also be able to help with the below:
I'm still having problems with the Selective Refresh.  As I said above, I have Notifications working, but the CSS is updating in a funny way.  When a change is made to the control (like modifying the background color), the change flashes for a moment, then the previously saved version of the CSS goes back into the <style id="wp-custom-css"> element.

I'm thinking that I need to tell the Setting class to send something to the preview, like by extending `WP_Customize_Setting::preview()`, or something to that effect.  

Any ideas?

src/wp-includes/class-wp-custom-css.php
```diff
@@ -0,0 +1,284 @@
...
+ * @since 4.7.0
+ */
+class WP_Custom_CSS {
+
+	/**
+	 * The Style CPT slug.
+	 *
+	 * Used as a constant here to avoid duplicating code.
+	 */
+	const POST_TYPE_SLUG = 'wp_custom_css';
```
celloexpressions: How badly could this break plugins if we don't prefix it? I don't think most of the other core post types have `wp_` prefixes.

src/wp-includes/class-wp-custom-css.php
```diff
@@ -0,0 +1,284 @@
...
+				'name' => __( 'WP Custom CSS' ),
+				'singular_name' => __( 'WP Custom CSS' ),
+				'edit_item' => __( 'Edit WP Custom CSS' ),
+				'add_new_item' => __( 'Add New WP Custom CSS' ),
+			),
+			'public'           => true,
+			'show_ui'          => false,
+			'hierarchical'     => false,
+			'rewrite'          => false,
+			'menu_position'    => 50,
```
celloexpressions: I think we can remove this line since UI is off.

src/wp-includes/class-wp-custom-css.php
```diff
@@ -0,0 +1,284 @@
...
+			'delete_with_user' => true,
+			'_builtin'         => true, /* internal use only. don't use this when registering your own post type. */
+			'supports'         => array( 'title', 'revisions' ),
+			'capabilities'     => array(
+				'delete_posts'           => 'edit_theme_options',
+				'delete_post'            => 'edit_theme_options',
+				'delete_published_posts' => 'edit_theme_options',
+				'delete_private_posts'   => 'edit_theme_options',
+				'delete_others_posts'    => 'edit_theme_options',
+				'edit_post'              => 'edit_theme_options',
```
celloexpressions: At least all of the edit caps should be `unfiltered_css`; might make sense to use that for all of them, but theme options also makes sense.

src/wp-includes/class-wp-custom-css.php
```diff
@@ -0,0 +1,284 @@
...
+	}
+
+	/**
+	 * Ensure Style posts exist for all installed themes
+	 *
+	 * @since 4.7.0
+	 *
+	 * @return bool
+	 */
+	public function maybe_create_posts() {
```
celloexpressions: Would it be better to only create the post when the theme is first previewed in the customizer, or once there's CSS? Guessing there are reasons to do it this way, though.

celloexpressions: Larger multisite networks might have a lot of themes installed that won't ever be used on a given site, that might be one reason to take a more conservative approach here.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -2295,6 +2298,31 @@ public function register_controls() {
...
+		) );
+
+		$this->add_setting( new WP_Customize_Custom_CSS_Setting( $this, 'wp_custom_css', array(
+			'capability' => 'unfiltered_css',
+		) ) );
+
+		$this->add_control( new WP_Customize_Code_Editor_Control( $this, 'wp_custom_css', array(
+			'type'        => 'code_editor',
+			'section'     => 'wp_custom_css',
+			'settings'    => 'wp_custom_css',
```
celloexpressions: This line can be removed, since the setting id matches the control id.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -2295,6 +2298,31 @@ public function register_controls() {
 			'section' => 'static_front_page',
 			'type' => 'dropdown-pages',
 		) );
+
+		/* Custom CSS */
+		$this->add_section( 'wp_custom_css', array(
```
celloexpressions: We don't need to prefix the section; it'll merge with any added by themes and plugins, and theirs will be removed over time.

src/wp-includes/class-wp-customize-section.php
```diff
@@ -344,13 +356,13 @@ public function print_template() {
 	 */
 	protected function render_template() {
 		?>
-		<li id="accordion-section-{{ data.id }}" class="accordion-section control-section control-section-{{ data.type }}">
+		<li id="accordion-section-{{ data.id }}" class="accordion-section control-section control-section-{{ data.type }} <# if ( data.description_hidden ) { #>customize-description-hidden<# } #>">
```
celloexpressions: We should be able to do this without adding a class here.

src/wp-includes/class-wp-customize-section.php
```diff
@@ -344,13 +356,13 @@ public function print_template() {
...
 			<h3 class="accordion-section-title" tabindex="0">
 				{{ data.title }}
 				<span class="screen-reader-text"><?php _e( 'Press return or enter to open this section' ); ?></span>
 			</h3>
 			<ul class="accordion-section-content">
-				<li class="customize-section-description-container">
+				<li class="customize-section-description-container section-meta">
```
celloexpressions: I'm not sure why this doesn't have the `customize-info` class already, but can we see if things explode if we use that instead of `section-meta`?

src/wp-includes/class-wp-customize-section.php
```diff
@@ -361,8 +373,15 @@ protected function render_template() {
 							</span>
 							{{ data.title }}
 						</h3>
+						<# if ( data.description && data.description_hidden ) { #>
+							<button class="customize-help-toggle dashicons dashicons-editor-help" tabindex="0" aria-expanded="false"><span class="screen-reader-text"><?php _e( 'Help' ); ?></span></button>
```
celloexpressions: The `<button>` element needs `type="button"`, and the tabindex can be removed.

celloexpressions: (this allows the extra JS events on keydown and preventDefault to be removed and uses native tabindex)

src/wp-includes/customize/class-wp-customize-code-editor-control.php
```diff
@@ -0,0 +1,83 @@
...
+	public $type = 'code_editor';
+
+	/**
+	 * Enqueue scripts and styles for this Control.
+	 *
+	 * @since 4.7.0
+	 * @access public
+	 */
+	public function enqueue() {
+		wp_enqueue_style( 'customize-code-editor' );
```
celloexpressions: This should be part of customize-controls.css.

src/wp-includes/customize/class-wp-customize-code-editor-control.php
```diff
@@ -0,0 +1,83 @@
...
+		<label>
+			<# if ( data.label ) { #>
+				<span class="customize-control-title">{{{ data.label }}}</span>
+			<# } #>
+			<# if ( data.description ) { #>
+				<span class="description customize-control-description">{{{ data.description }}}</span>
+			<# } #>
+			<div class="customize-control-content">
+				<div class="customize-control-code-editor-line-numbers">
+					<?php foreach ( range( 1, 999 ) as $i ) {
```
celloexpressions: I wonder if there's a trick for magically doing this with CSS or JS? If not, this works.

src/wp-includes/customize/class-wp-customize-custom-css-setting.php
```diff
@@ -0,0 +1,275 @@
...
+	 * @return mixed true|WP_Error True if the input was validated, otherwise WP_Error.
+	 */
+	public function validate_css( $validity, $css ) {
+		// Make sure that there is a closing brace for each opening brace.
+		if ( ! self::validate_balanced_characters( '{', '}', $css ) ) {
+			$validity->add( 'unbalanced_braces', __( 'Your braces <code>{}</code> are unbalanced. Make sure there is a closing <code>}</code> for every opening <code>{</code>.' ) );
+		}
+
+		// Ensure brackets are balanced.
+		if ( ! self::validate_balanced_characters( '[', ']', $css ) ) {
```
celloexpressions: I suppose we might as well check for balanced parentheses while we're at it, since it'll be the same check as braces and brackets (used for rgba colors, calc(), etc.).

src/wp-includes/class-wp-custom-css.php
westonruter: I recommend removing this entirely. Only create the post when there is CSS actually being `save`ed.

src/wp-includes/customize/class-wp-customize-custom-css-setting.php
```diff
@@ -0,0 +1,275 @@
...
+			return false;
+		}
+
+		$style_post = get_post( $current_theme_post_id );
+
+		if ( empty( $style_post ) ) {
+			return false;
+		}
+
+		$args['ID'] = $current_theme_post_id;
```
westonruter: As I commented elsewhere, I suggest that the post only be created as needed. So `$current_theme_post_id` should be empty in that case and `wp_insert_post()` should be used instead.

src/wp-includes/customize/class-wp-customize-custom-css-setting.php
```diff
@@ -0,0 +1,275 @@
...
+	 *
+	 * @param WP_Customize_Manager $manager The Customize Manager class.
+	 * @param string               $id      An specific ID of the setting. Can be a
+	 *                                      theme mod or option name.
+	 * @param array                $args    Setting arguments.
+	 */
+	public function __construct( $manager, $id, $args = array() ) {
+		parent::__construct( $manager, $id, $args = array() );
+
+		add_filter( "customize_value_{$this->id_data['base']}", array( $this, 'get_value' ) );
```
westonruter: This filter could get added multiple times if more than one instance of `WP_Customize_Custom_CSS_Setting` is made. You may want to let `get_value` be `static` and only `add_filter` if not `has_filter`.

src/wp-includes/customize/class-wp-customize-custom-css-setting.php
```diff
@@ -0,0 +1,275 @@
...
+		add_filter( "customize_validate_{$this->id}", array( $this, 'validate_css' ), 10, 2 );
+		add_filter( "customize_sanitize_{$this->id}", array( $this, 'sanitize_css' ), 10, 2 );
+		add_action( "customize_update_{$this->type}", array( $this, 'update_setting' ) );
```
westonruter: As with above, this action can get triggered multiple times if more than one setting instance is made (which isn't what is being done in the patch, but still a plugin could do it). So again, as above, might want to use a `static` method instead and only `add_action` if not `has_action`.

src/wp-includes/customize/class-wp-customize-custom-css-setting.php
```diff
@@ -0,0 +1,275 @@
...
+	 *
+	 * @filter customize_value_{$this->id_data['base']}
+	 *
+	 * @since 4.7.0
+	 *
+	 * @param string $value The value.
+	 *
+	 * @return string
+	 */
+	public function get_value( $value ) {
```
westonruter: Consider `static`.

You'll need to figure out which `$setting` instance is being use by looking at the second `$setting` argument.

src/wp-includes/customize/class-wp-customize-custom-css-setting.php
```diff
@@ -0,0 +1,275 @@
...
+		$curr_style_post_id = WP_Custom_CSS::get_style_post_id();
```
westonruter: You'll need to look if there is any value being provided in the customized state, and if so, return it instead of the saved value in the DB.

So like this:

```php
$value = $setting->post_value();
if ( ! is_null( $value ) ) {
    return $value;
}
```

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -2295,6 +2298,31 @@ public function register_controls() {
...
+			'description_hidden' => true,
+			'description' => sprintf( '%s<br /><a href="%s" class="external-link" target="_blank">%s&nbsp;<span class="screen-reader-text">%s</span></a>',
+				__( 'CSS allows you to customize the appearance and layout of your site with code. Separate CSS is saved for each of your themes.' ),
+				'https://codex.wordpress.org/Know_Your_Sources#CSS',
+				__( 'Learn more about CSS' ),
+				__( '(link opens in a new window)' )
+			),
+		) );
+
+		$this->add_setting( new WP_Customize_Custom_CSS_Setting( $this, 'wp_custom_css', array(
```
westonruter: The setting ID might need to incorporate the theme name so that when previewing multiple themes and modifying the CSS in each, the CSS modifications for each will get separately saved. This is currently being looked at in changesets, where if you have multiple `theme_mod` settings that you modify in one theme, and then you try previewing a different theme in the same customizer session, you can modify the theme mods in the other theme and they won't clobber the values you provided for the first theme if you switch back.

src/wp-includes/js/customize-preview.js
```diff
@@ -257,6 +257,12 @@
 			$( 'body' ).toggleClass( 'wp-custom-logo', !! setting.get() );
 			setting.bind( function( attachmentId ) {
 				$( 'body' ).toggleClass( 'wp-custom-logo', !! attachmentId );
+			});
+		});
+
+		api( 'wp_custom_css', function( value ) {
+			value.bind( function( to ) {
+				$( '#wp-custom-css' ).html( to );
```
westonruter: I think this should use `text()` instead of `html()`.

src/wp-includes/customize/class-wp-customize-custom-css-setting.php
westonruter: This will also ensure that the custom CSS is printed in the document coming from PHP, as opposed to assuming that the value will get synced from the pane via `postMessage`. With changesets you can pop-out the preview into a separate window without the pane at all, and even preview in incognito, and in this case you'd absolutely need to ensure that the `value` method is returning the previewed value as opposed to the saved value in the DB.

src/wp-includes/theme.php
```diff
@@ -1401,6 +1401,35 @@ function _custom_background_cb() {
...
+ *
+ * @todo sanitize this.
+ *
+ * @since 4.7.0
+ */
+function wp_custom_css_cb() {
+	$styles = wp_get_custom_css();
+	if ( $styles || is_customize_preview() ) : ?>
+		<style type="text/css" id="wp-custom-css">
+			<?php echo $styles; ?>
```
westonruter: Let this be `<?php echo esc_html( $styles ); ?>`. We wouldn't want a user entering “CSS” looking like:

```
body{color:red}</style><script>alert('EVIL)</script>
```

src/wp-includes/class-wp-custom-css.php
```diff
@@ -0,0 +1,284 @@
...
+				'edit_item' => __( 'Edit WP Custom CSS' ),
+				'add_new_item' => __( 'Add New WP Custom CSS' ),
+			),
+			'public'           => true,
+			'show_ui'          => false,
+			'hierarchical'     => false,
+			'rewrite'          => false,
+			'menu_position'    => 50,
+			'query_var'        => false,
+			'delete_with_user' => true,
```
westonruter: I think this should actually be `false`. Also, add `'can_export' => true`.

src/wp-includes/customize/class-wp-customize-custom-css-setting.php
johnregan3: @westonruter So, if the $setting->post_value() is not found, then get the post from the DB?

westonruter: @johnregan3 Exactly. The presence of a post (changeset) value means that the CSS is modified in the customized state, and so the DB value needs to be overridden. Oh, but an important clarification there. This should only happen if `preview()` was called. So you should add another action for `customize_preview_{$this->id}` and inside the function callback do `$this->is_previewing = true`. Then you can start out the `get_value()` method with this revised snippet:

```php
$value = $setting->post_value();
if ( $this->is_previewing && ! is_null( $value ) ) {
    return $value;
}
```

src/wp-admin/js/customize-controls.js
```diff
@@ -1589,10 +1586,15 @@
 			nodes  = control.container.find('[data-customize-setting-link]');
 			radios = {};
 
+			console.log( nodes.prevObject );
```
johnregan3: @westonruter At this point, I get (basically) `li#customize-control-wp_custom_css.customize-control.customize-control-code_editor` in the console, which lets me know that the node is being detected.  

However, on line 1596 below, the textarea (`this`) does not appear, although input and select nodes do.

src/wp-admin/js/customize-controls.js
```diff
@@ -2930,6 +2956,30 @@
...
+		/**
+		 * @inheritdoc
+		 */
+		ready: function() {
+			var control = this, textarea;
+			textarea = control.container.find( '.customize-control-code-editor-textarea' ).get( 0 );
+
+			// Scroll the line numbers with the textarea.
+			$( textarea ).on( 'scroll resize', function () {
+				$( textarea ).parents( '.customize-control-content' ).find( '.customize-control-code-editor-line-numbers' ).scrollTop( $( this ).scrollTop() );
```
westonruter: This `$( textarea ).parents( '.customize-control-content' ).find( '.customize-control-code-editor-line-numbers' )` should be stored in a variable outside this handler. It doesn't need to be re-queried with each event.

src/wp-includes/theme.php
```diff
@@ -1401,6 +1401,115 @@ function _custom_background_cb() {
...
+ *
+ * @return string The Custom CSS Post content.
+ */
+function wp_get_custom_css( $theme_name = '' ) {
+	$custom_css_post = wp_get_custom_css_by_theme_name( $theme_name );
+	if ( empty( $custom_css_post->post_content ) ) {
+		return '';
+	}
+
+	return $custom_css_post->post_content;
```
westonruter: Per Helen [in Slack](https://wordpress.slack.com/archives/core/p1476305824005800):

> I think it’d also be good to address output and possibilities for extending/altering output.

I think we'll need a filter here.

src/wp-includes/customize/class-wp-customize-code-editor-control.php
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+/**
+ * Class WP_Customize_Code_Editor_Control
+ *
+ * @since 4.7.0
+ *
+ * @see WP_Customize_Control
+ */
+class WP_Customize_Code_Editor_Control extends WP_Customize_Control {
```
aaroncampbell: I'd rather this start simpler as well. Use a regular textarea and iterate from there.

src/wp-includes/theme.php
```diff
@@ -1401,6 +1401,115 @@ function _custom_background_cb() {
...
+ * Also, if the Custom CSS post doesn't exist, then create it.
+ *
+ * @since 4.7.0
+ *
+ * @param string $theme_name Optional. A Theme object stylesheet name. Defaults to the current theme.
+ * @param string $value Optional. A value to update the post content.
+ *
+ * @return object WP_Post Object
+ */
+function wp_get_custom_css_by_theme_name( $theme_name = '', $value = '' ) {
```
johnregan3: That makes sense.  I'll get on that.

src/wp-includes/theme.php
```diff
@@ -1401,6 +1401,115 @@ function _custom_background_cb() {
...
+	 * use the post_content.
+	 *
+	 * This is used specifically in wp_get_custom_css().
+	 */
+	if ( empty( $value ) && ! empty( $post->post_content ) ) {
+		$args['post_content'] = $post->post_content;
+	}
+
+	// Create/update the post.
+	$post_id = wp_insert_post( wp_slash( $args ) );
```
westonruter: You'll need to make sure you remove the kses filters here so that it doesn't corrupt the CSS. See https://github.com/xwp/wordpress-develop/blob/4fa326d5851c27837ad31bab0926282b4f3d59bf/src/wp-includes/class-wp-customize-manager.php#L1716-L1731

src/wp-includes/theme.php
```diff
@@ -1401,6 +1401,115 @@ function _custom_background_cb() {
...
+	 * If a new $value is passed, use it as the post_content.
+	 */
+	$args = array(
+		'post_content' => $value,
+		'post_type'    => 'custom_css',
+		'post_title'   => $theme_name,
+		'post_status'  => 'publish',
+	);
+
+	$post = get_page_by_title( $theme_name, 'OBJECT', 'custom_css' );
```
westonruter: This should be `OBJECT` constant not `'OBJECT'` string.

johnregan3:
@celloexpressions @westonruter 

This is ready for another review pass, so that if you have some time this weekend we can get a head start on next week.  I've updated this branch quite a bit, and and I think it cleaned up the feature very nicely.

- [x] Removed Code Editor control
- [x] Simplified the Section help icon usage greatly
- [x] Renamed `wp_custom_css` to `custom_css` throughout per @celloexpressions 's request
- [x] Use the theme name in the Setting ID
- [x] All CSS now resides inside `customize-controls.css`
- [x] Implemented `has_filter()` in the Setting per @westonruter 
- [x] For `get_value()`, we now check for a customized state
- [x] Moved the CPT into `create_initial_post_types()`
- [x] Completely removed that crazy post creation bit.  Now it's only created in `update_setting()`
- [x] Several other bits and bobs mentioned in your reviews

I still have these things to do:
  - [ ]  Save the Post ID into a theme_mod
  - [ ]  Add a filter to the CSS output
  - [ ]  Work out removing those string literals when the CSS is validated.  (this could be tricky)
  - [ ]  Maybe move the icon description toggle into its own function so it can be used by both Panels and Sections

If this is not a complete list, please let me know so I don't miss anything.

src/wp-admin/css/customize-controls.css
```diff
@@ -53,6 +53,10 @@ body {
 	margin-bottom: 15px;
 }
 
+#customize-controls .customize-info.section-meta {
+	margin-bottom: 10px;
```
celloexpressions: Typical spacing between section headings is 15px.

src/wp-admin/css/customize-controls.css
```diff
@@ -124,22 +128,30 @@ body {
...
 	border-top: 1px solid #ddd;
 }
+
 #customize-controls .customize-info .customize-panel-description.open + .no-widget-areas-rendered-notice {
 	border-top: none;
 }
 
-#customize-controls .customize-info .customize-panel-description p:first-child {
+#customize-controls .customize-info .customize-section-description {
+	margin-bottom: 10px;
```
celloexpressions: Should be 15px

src/wp-admin/js/customize-controls.js
```diff
@@ -746,6 +746,26 @@
...
+			// This is very similar to what is found for api.Panel.attachEvents().
+			section.container.find( '.customize-section-title .customize-help-toggle' ).on( 'click', function() {
+
+				meta = section.container.find( '.section-meta' );
+				if ( meta.hasClass( 'cannot-expand' ) ) {
+					return;
+				}
+
+				content = meta.find( '.customize-section-description:first' );
+				if ( meta.hasClass( 'open' ) ) {
```
celloexpressions: These could be condensed by using `toggleClass`, `slideToggle`, and:
```
.attr('aria-expanded', function ( i, attr ) {
	return attr === 'true' ? 'false' : 'true';
});
```

src/wp-includes/theme.php
```diff
@@ -1427,13 +1427,21 @@ function wp_custom_css_cb() {
...
 
-	return $custom_css_post->post_content;
+	/**
+	 * Modify the Custom CSS Output into the <head>.
+	 *
+	 * @since 4.7.0
+	 *
+	 * @param string $css CSS pulled in from the Custom CSS CPT.
+	 */
+	return apply_filters( 'custom_css_output', $css );
```
westonruter: Perhaps `wp_get_custom_css` instead. Compare with the `get_custom_logo` filter.

johnregan3:
In my latest work, I've done the following:

- [x]  PHPUnit Tests for `WP_Customize_Custom_CSS_Setting`
- [x]  Changed Margin around Section title as requested by @celloexpressions 
- [x] Updated Help Icon open/close JS, per @celloexpressions 
- [x] Implemented a secondary check for `content:` if a validation error in balanced characters is found. (this is in lieu of removing those string literals for the time being)

I've opted to _not_ move the help icon toggling into a new function as I don't really have time to implement/test it.

src/wp-includes/theme.php
```diff
@@ -1401,6 +1401,90 @@ function _custom_background_cb() {
...
+	}
+
+	$post = null;
+	$post_id = null;
+	if ( get_stylesheet() === $stylesheet ) {
+		$post_id = get_theme_mod( 'custom_css_post_id' );
+		if ( ! $post_id ) {
+
+			// Cache value in the theme mod to eliminate a query with each request.
+			$post = get_page_by_title( $stylesheet, OBJECT, 'custom_css' ); // @todo This needs to look up by post_name instead since it is indexed.
```
westonruter: We should look-up by `post_name` instead of `post_title` since the former is indexed. Changesets provides an example of this.

src/wp-includes/theme.php
```diff
@@ -1401,6 +1401,90 @@ function _custom_background_cb() {
...
+			set_theme_mod( 'custom_css_post_id', $post ? $post->ID : -1 );
+		} elseif ( $post_id > 0 ) {
+			$post = get_post( $post_id );
+		}
+	} else {
+		// Theme is not active, so the cached post ID cannot be sourced from the theme mods.
+		$post = get_page_by_title( $stylesheet, OBJECT, 'custom_css' ); // @todo This needs to look up by post_name instead since it is indexed.
```
westonruter: Ditto. We should look-up by `post_name` instead of `post_title` since the former is indexed. Changesets provides an example of this.

src/wp-includes/customize/class-wp-customize-custom-css-setting.php
```diff
@@ -0,0 +1,389 @@
...
+	 * @access public
+	 * @since 4.7.0
+	 *
+	 * @todo Is this necessary anymore?
+	 *
+	 * @param string $css The input string.
+	 * @return string Limited sanitized CSS.
+	 */
+	public function sanitize( $css ) {
+		return parent::sanitize( wp_sanitize_css( $css ) );
```
westonruter: Does `wp_sanitize_css()` provide any value now that we have `unfiltered_css`? What does it sanitize for us? If it helps, then we need unit tests to show what it does.

src/wp-includes/customize/class-wp-customize-custom-css-setting.php
```diff
@@ -0,0 +1,389 @@
...
+	 * @return int|false The post ID or false if the value could not be saved.
+	 */
+	public function update( $value ) {
+		/*
+		 * If the theme associated with this setting is not the loaded theme
+		 * then the update must short-circuit because theme mods arne't set
+		 * for non-active themes.
+		 */
+		if ( $this->manager->get_stylesheet() !== $this->stylesheet ) {
+			return false;
```
westonruter: Actually we can remove this conditional. We can just only do `set_theme_mod()` if the stylesheet matches.

westonruter:
Committed to core in https://core.trac.wordpress.org/changeset/38829

PR #155, opened 2016-09-08
WIP: Make panel headers sticky (#34343)

delawski:


westonruter:
We may be able to avoid adding new events for `expandPanel` and `collapsePanel` because there are already state changes we can listen to on the panels themselves. There is `panel.expanded` which is a `Value` that can be listened to:

```js
panel.expanded.bind( function( isExpanded ) {
    // ...
} );
```

delawski:
@westonruter - I have been working on the update and I'm not sure which way is better. You suggested to bind to `panel.expanded` Value changes. However in such case, I would have to bind to each and every panel/section, as I need to operate on a particular container being expanded/collapsed.

I thought it might be better to bind to jQuery custom events (`expanded` and `collapsed`) and check each panel/section state by iterating through `api.panel` and `api.section`, as it is done in other parts of the code? In my WIP I used that approach for now.

Do you maybe see any other, cleaner and smarter way to solve that issue? I was thinking of passing panel/section object through `data` object with `expanded` and `collapsed` jQuery events, but I've noticed those events are triggered multiple times in a row (btw. is it intentional? I used `_.debounce()` to prevent overcome the issue).

westonruter:
@delawski I guess I tend toward using the `expanded` state to listen for changes rather than looking at the `extended` event because that's what I've always used. I've not used the `extended`  or `collapsed` jQuery events. You're now more experienced with how the panels are architected, so if you feel that these are a better option, then you probably know better. 

> but I've noticed those events are triggered multiple times in a row (btw. is it intentional? I used _.debounce() to prevent overcome the issue).

It's not intentional, but I can see why it is happening. The `completeCallback` is getting called even in the case of `unchanged`, and this callback is including the triggering of these events. It should probably be improved to not trigger those events in the case of `unchanged`.

PR #157, opened 2016-09-23
[Alternate PR] Trac 38021  : Remove Customizer support for IE8

kienstra:
**Request To Review New Pull Request**

Hi @westonruter,
Thanks for [catching](https://github.com/xwp/wordpress-develop/pull/156#discussion_r79984932) the issue with the `no-customize-support` being replaced with `no-no-customize-support` in the other [pull request](https://github.com/xwp/wordpress-develop/pull/156#discussion_r79984932). I opened this new pull request, because the other has a merge conflict after its [first commit](https://github.com/xwp/wordpress-develop/pull/156/commits/0e873bdcab65bd84e7af0b294e5c4433cd708ed0) went into Core.

In the [script that removes IE8 support](https://github.com/xwp/wordpress-develop/compare/trac-38021-ie8?expand=1#diff-c3492b4a8b647a363b5d0de02d8a39c8R2161), this PR [removes](https://github.com/xwp/wordpress-develop/compare/trac-38021-ie8?expand=1#diff-c3492b4a8b647a363b5d0de02d8a39c8R2163) either `.customize-support` or  `.no-customize-support` from `<body>`, before adding `.no-customize-support`. I don’t think we can be sure which, if any class will appear.

When the admin header bar is present on the front end of the site, `<body>` will [usually have](https://github.com/xwp/wordpress-develop/blob/trac-38021-ie8/src/wp-includes/post-template.php#L711) the `no-customize-support` class before this script runs. So IE8 support would already be removed, based on the conditional comments in these commits.

But this assumes:

* The theme calls [body_class()](https://github.com/xwp/wordpress-develop/blob/trac-38021-ie8/src/wp-includes/post-template.php#L555) in `<body>`
* No change of the `no-customize-support` class in the filter [body_class](https://github.com/xwp/wordpress-develop/blob/trac-38021-ie8/src/wp-includes/post-template.php#L766)

So this commit [removes either class](https://github.com/xwp/wordpress-develop/compare/trac-38021-ie8?expand=1#diff-c3492b4a8b647a363b5d0de02d8a39c8R2163): `customize-support` or `no-customize-support`:

```document.body.className = document.body.className.replace( /(^|\s)(no-)?customize-support/, '' ).trim() + ' no-customize-support';```

This seems to be the same approach that the [original script takes](https://github.com/xwp/wordpress-develop/compare/trac-38021-ie8?expand=1#diff-c3492b4a8b647a363b5d0de02d8a39c8R2177). Substituting variable values, this would be:
`document.body[ ‘className’ ] = document.body[ ‘className’ ].replace( new RegExp('(^|\\s+)(no-)?customize-support(\\s+|$)’), ‘ ‘ );`

The [original script](https://github.com/xwp/wordpress-develop/compare/trac-38021-ie8?expand=1#diff-c3492b4a8b647a363b5d0de02d8a39c8R2166) is wrapped in the [downlevel conditional](https://msdn.microsoft.com/en-us/library/ms537512(v=vs.85).aspx#dlreveale) `<![if gte IE 9]>`. So it will run on all non-IE browser, and IE9+. I couldn’t find an exact equivalent to an `else` statement in conditional comments.

The new IE8 script is wrapped in a more normal conditional comment, which only IE will recognize. Luckily, non-IE browsers will parse it as a comment, and won’t run its inefficient Regex and string concatenation.

### DIFF
--- a/src/wp-includes/theme.php
+++ b/src/wp-includes/theme.php
@@ -2169,6 +2169,7 @@ function wp_customize_support_script() {
 <?php		endif; ?>
 
 			b[c] = b[c].replace( rcs, ' ' );
+			// The customizer requires postMessage and CORS (if the site is cross domain)
 			b[c] += ( window.postMessage && request ? ' ' : ' no-' ) + cs;
 		}());
 	</script>
--- a/src/wp-includes/theme.php
+++ b/src/wp-includes/theme.php
@@ -2158,25 +2158,28 @@ function wp_customize_support_script() {
 	$cross_domain = ( strtolower( $admin_origin[ 'host' ] ) != strtolower( $home_origin[ 'host' ] ) );
 
 	?>
-	<script type="text/javascript">
-		(function() {
-			var request, b = document.body, c = 'className', cs = 'customize-support', rcs = new RegExp('(^|\\s+)(no-)?'+cs+'(\\s+|$)');
+	<!--[if lte IE 8]>
+		<script type="text/javascript">
+			document.body.className = document.body.className.replace( /(^|\s)(no-)?customize-support/, '' ).trim() + ' no-customize-support';
+		</script>
+	<![endif]-->
+	<![if gte IE 9]>
+		<script type="text/javascript">
+			(function() {
+				var request, b = document.body, c = 'className', cs = 'customize-support', rcs = new RegExp('(^|\\s+)(no-)?'+cs+'(\\s+|$)');
 
-<?php		if ( $cross_domain ): ?>
-			request = (function(){ var xhr = new XMLHttpRequest(); return ('withCredentials' in xhr); })();
-<?php		else: ?>
-			request = true;
-<?php		endif; ?>
+		<?php	if ( $cross_domain ) : ?>
+				request = (function(){ var xhr = new XMLHttpRequest(); return ('withCredentials' in xhr); })();
+		<?php	else : ?>
+				request = true;
+		<?php	endif; ?>
 
-			b[c] = b[c].replace( rcs, ' ' );
-			// The customizer requires postMessage and CORS (if the site is cross domain)
-			b[c] += ( window.postMessage && request ? ' ' : ' no-' ) + cs;
-		}());
-	</script>
-
-	<!--[if lte IE 8]><script type="text/javascript">
-		document.body.className = document.body.className.replace( 'customize-support', 'no-customize-support' );
-	</script><![endif]-->
+				b[c] = b[c].replace( rcs, ' ' );
+				// The customizer requires postMessage and CORS (if the site is cross domain)
+				b[c] += ( window.postMessage && request ? ' ' : ' no-' ) + cs;
+			}());
+		</script>
+	<![endif]>
 	<?php
 }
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/wp-includes/theme.php [hunk 1]
```diff
@@ -2158,20 +2158,28 @@ function wp_customize_support_script() {
...
-			b[c] = b[c].replace( rcs, ' ' );
-			b[c] += ( window.postMessage && request && Array.prototype.indexOf ? ' ' : ' no-' ) + cs;
-		}());
-	</script>
+	<!--[if lte IE 8]>
+		<script type="text/javascript">
+			document.body.className = document.body.className.replace( /(^|\s)(no-)?customize-support/, '' ).trim() + ' no-customize-support';
+		</script>
+	<![endif]-->
+	<![if gte IE 9]>
```
westonruter: I think this should be `<!--<![if gte IE 9]> <!-- -->` so that the text won't get leaked to non-IE browsers.

See http://stackoverflow.com/a/10159787/93579

src/wp-includes/theme.php
```diff
@@ -2158,20 +2158,28 @@ function wp_customize_support_script() {
...
+		<?php	else : ?>
+				request = true;
+		<?php	endif; ?>
+
+				b[c] = b[c].replace( rcs, ' ' );
+				// The customizer requires postMessage and CORS (if the site is cross domain)
+				b[c] += ( window.postMessage && request ? ' ' : ' no-' ) + cs;
+			}());
+		</script>
+	<![endif]>
```
westonruter: And per above, I think this should be `<!-- <![endif]-->`

src/wp-includes/theme.php
```diff
@@ -2158,20 +2158,28 @@ function wp_customize_support_script() {
...
-			request = true;
-<?php		endif; ?>
-
-			b[c] = b[c].replace( rcs, ' ' );
-			b[c] += ( window.postMessage && request && Array.prototype.indexOf ? ' ' : ' no-' ) + cs;
-		}());
-	</script>
+	<!--[if lte IE 8]>
+		<script type="text/javascript">
+			document.body.className = document.body.className.replace( /(^|\s)(no-)?customize-support/, '' ).trim() + ' no-customize-support';
```
westonruter: `String.prototype.trim()` is not supported in IE8 😄: http://caniuse.com/#search=trim

You can just remove the `trim` entirely because it's no problem to have extra whitespace in a CSS `className`.

westonruter: Also, for good measure, the regex should be `/(^|\s)(no-)?customize-support(?=\s|$)/`

kienstra: Thanks, @westonruter. This [commit](https://github.com/xwp/wordpress-develop/pull/157/commits/e85710edb1dceca03589463612f8deb96d79d90f) changes the regex.

westonruter:
@kienstra I don't have IE8 handy. You can confirm that the Customize links are indeed hidden in IE8?

kienstra:
Hi @westonruter,

Yes, the Customize links are hidden in IE8. 😄 Here's a [screenshot](https://cldup.com/hxoMwLimwR.jpg) of the link hidden from the admin bar on the front-end. And [here's one](https://cldup.com/BgROvvi0GB.png) of the "Customize" link not appearing in the `/wp-admin` "Appearance" section.

PR #161, opened 2016-10-02
Trac 30937: Implement Customize Changesets (née Transactions aka Snapshots)

westonruter:
Supersedes #61.

For outstanding issues see project: https://github.com/xwp/wordpress-develop/projects/1

See https://core.trac.wordpress.org/ticket/30937#comment:56


src/wp-includes/js/customize-preview.js
```diff
@@ -37,73 +37,348 @@
...
-				/*
-				 * Note the shift key is checked so shift+click on widgets or
-				 * nav menu items can just result on focusing on the corresponding
-				 * control instead of also navigating to the URL linked to.
-				 */
-				if ( event.shiftKey || isInternalJumpLink ) {
-					return;
-				}
-				self.send( 'scroll', 0 );
-				self.send( 'url', link.prop( 'href' ) );
```
westonruter: Actually, I think this needs to be restored. For compatibility with plugins that filter `wp.customize.previewer.previewUrl()` to add additional constraints as to what can be previewed and also to inject additional query parameters into requests, we should continue to use that method if possible.

src/wp-includes/functions.php
```diff
@@ -5529,3 +5529,20 @@ function wp_raise_memory_limit( $context = 'admin' ) {
...
 }
+
+/**
+ * Generate a random UUID (version 4).
+ *
+ * @since 4.7.0
+ *
+ * @return string UUID.
+ */
+function wp_generate_uuid4() {
```
westonruter: I guess, at the moment, there is no invocation of this function outside the manager. But plugins may very well want to call this if they create changeset posts programmatically.

aaroncampbell: It looks like there's a comment in `_wp_customize_changeset_filter_insert_post_data()` referencing the possible need for this? It seems like if there's a possible use case like that, then plugins might need it as well. It doesn't seem particularly tied to `WP_Customize_Manager` either, since it's just a spec function, right?

src/wp-admin/js/customize-controls.js
```diff
@@ -64,6 +81,92 @@
...
+				requestDeferred.reject( data );
+			} );
+			request.always( function( data ) {
+				if ( data.setting_validities ) {
+					api._handleSettingValidities( {
+						settingValidities: data.setting_validities
+					} );
+				}
+			} );
+		}, api.previewer.refreshBuffer ); // Let this come from api.settings.updateChangesetBuffer.
```
westonruter: We should have a `wp.customize.settings.timeoutIntervals` (or something) which this can read from.

src/wp-admin/js/customize-controls.js
```diff
@@ -2989,142 +3117,81 @@
...
-						if ( ! isDynamicallyCreated || ! _.isUndefined( activeConstructs[ id ] ) ) {
-							if ( activeConstructs[ id ] ) {
-								construct.activate();
-							} else {
-								construct.deactivate();
-							}
-						}
-					} );
-				} );
+			// @todo Wait until 0 === processing?
```
westonruter: This can be removed because `processing` is being watched in `wp.customize.previewer.refresh()`.

src/wp-admin/js/customize-controls.js
```diff
@@ -3307,32 +3353,213 @@
...
 
 				// If we found a matching result, return it. If not, bail.
 				return result ? result : null;
 			});
 
-			// Refresh the preview when the URL is changed (but not yet).
-			this.previewUrl.bind( this.refresh );
+			// Change preview iframe URL when the previewUrl changes.
+			previewer.setIframeSrc = _.bind( previewer.setIframeSrc, previewer );
+			previewer.previewUrl.bind( previewer.setIframeSrc );
```
westonruter: This should wait until `processing` state is `0`, to ensure the changeset is populated.

src/wp-admin/js/customize-controls.js
```diff
@@ -2989,142 +3117,81 @@
...
-					self.login( deferred );
+				if ( 0 === error ) {
+					previewFrame.login( deferred );
 					return;
 				}
 
 				// Check for cheaters.
-				if ( '-1' === response ) {
-					deferred.rejectWith( self, [ 'cheatin' ] );
+				if ( -1 === error ) {
```
aaroncampbell: Did something change to make -1 numeric here, or should this still be treating it as a string like before (`if ( '-1' === error )`)?

src/wp-admin/js/customize-controls.js
```diff
@@ -2989,142 +3117,81 @@
...
-			if ( this._ready ) {
-				this.unbind( 'ready', this._ready );
+			if ( previewFrame._ready ) {
+				previewFrame.unbind( 'ready', previewFrame._ready );
 			}
 
-			this._ready = function() {
+			previewFrame._ready = function( data ) {
 				ready = true;
+				readyData = data;
```
westonruter: Actually it is used below in the iframe `load` event handler:

```js
			previewFrame.iframe.one( 'load', function() {
				loaded = true;

				if ( ready ) {
					deferred.resolveWith( previewFrame, [ readyData ] );
```

src/wp-admin/js/customize-controls.js
westonruter: Yes, good question. Formerly the customizer was fetching the preview URL via Ajax where the customizer response would respond with `wp_die( -1 )` which would then get sent back as a string. In this patch, however, the error message is being sent back via `postMessage` and so the integer data type is not lost: https://github.com/xwp/wordpress-develop/blob/c4b1abb1dea90a99c033bca2fc057b24f3fa1a01/src/wp-includes/class-wp-customize-manager.php#L406-L429

tests/phpunit/tests/customize/manager.php
```diff
@@ -91,6 +91,7 @@ function test_not_doing_ajax() {
 	 * @ticket 30988
 	 */
 	function test_unsanitized_post_values() {
+		wp_set_current_user( $this->factory()->user->create( array( 'role' => 'administrator' ) ) );
```
ocean90: I think the tests should reuse the admin user, see `wpSetUpBeforeClass()` and https://github.com/aaronjorbin/develop.wordpress/blob/87dc042623d4065acc0defa2aae1de004f5069ad/tests/phpunit/tests/post/revisions.php#L13

src/wp-includes/post.php
```diff
@@ -111,6 +111,51 @@ function create_initial_post_types() {
...
+			'not_found'          => __( 'No changesets found.' ),
+			'not_found_in_trash' => __( 'No changesets found in Trash.' ),
+		),
+		'public' => false,
+		'_builtin' => true, /* internal use only. don't use this when registering your own post type. */
+		'map_meta_cap' => true,
+		'hierarchical' => false,
+		'rewrite' => false,
+		'query_var' => false,
+		'can_export' => true,
```
ocean90: That should probably be `false` by default.

src/wp-includes/theme.php
```diff
@@ -2066,24 +2066,111 @@ function check_theme_switched() {
...
+ * @param WP_Post $changeset_post    Changeset post object.
+ */
+function _wp_customize_publish_changeset( $changeset_post_id, $changeset_post ) {
+	global $wp_customize;
+	if ( empty( $wp_customize ) ) {
+		require_once ABSPATH . WPINC . '/class-wp-customize-manager.php';
+		$wp_customize = new WP_Customize_Manager( $changeset_post->post_name );
+	}
+	if ( ! did_action( 'customize_register' ) ) {
+		do_action( 'customize_register', $wp_customize );
```
ocean90: Missing hook docs: `/** This filter is documented in /wp-includes/class-wp-customize-manager.php */`

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -1091,56 +1507,314 @@ public function save() {
...
+			// Temporarily stop previewing the theme to allow switch_themes() to operate properly.
 			$this->stop_previewing_theme();
 			switch_theme( $this->get_stylesheet() );
 			update_option( 'theme_switched_via_customizer', true );
 			$this->start_previewing_theme();
 		}
 
+		// Gather the data for wp_insert_post()/wp_update_post().
+		$post_array = array(
+			'post_content' => wp_json_encode( $data, JSON_UNESCAPED_SLASHES | JSON_PRETTY_PRINT ),
```
ocean90: `JSON_UNESCAPED_SLASHES` was introduced in PHP 5.4. See https://core.trac.wordpress.org/ticket/30139 for `JSON_PRETTY_PRINT`.

src/wp-admin/export.php
```diff
@@ -254,7 +254,7 @@ function export_date_options( $post_type = 'post' ) {
 	</li>
 </ul>
 
-<?php foreach ( get_post_types( array( '_builtin' => false, 'can_export' => true ), 'objects' ) as $post_type ) : ?>
+<?php foreach ( get_post_types( array( 'can_export' => true ), 'objects' ) as $post_type ) : if ( $post_type->_builtin && 'customize_changeset' !== $post_type->name ) { continue; } ?>
```
westonruter: This works but it's seems kinda hacky.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -993,16 +1324,38 @@ public function current_theme( $current_theme ) {
 	 * @see WP_Customize_Setting::validate()
 	 *
 	 * @param array $setting_values Mapping of setting IDs to values to validate and sanitize.
+	 * @param array $options {
+	 *     Options.
+	 *
+	 *     @var bool $validate_existence    Whether a setting's existence will be checked.
+	 *     @var bool $validate_capability Whether the setting capability will be checked.
```
ocean90: `@var` should be `@type`. (https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#1-1-parameters-that-are-arrays)

In line 1330 is the description not aligned.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -1091,56 +1543,363 @@ public function save() {
...
 			switch_theme( $this->get_stylesheet() );
 			update_option( 'theme_switched_via_customizer', true );
 			$this->start_previewing_theme();
 		}
 
+		// Gather the data for wp_insert_post()/wp_update_post().
+		$json_options = 0;
+		if ( defined( 'JSON_UNESCAPED_SLASHES' ) ) {
+			$json_options |= JSON_UNESCAPED_SLASHES; // Introduced in PHP 5.4.
+		}
```
ocean90: Why is it OK to escape slashes in PHP < 5.3? Why doesn't it affect the decoding? What am I missing here? 🤔

westonruter: JSON allows `/` to be slashed or unslashed. As with the pretty printing option, this is purely for improved readability when looking at the raw JSON. Normally `json_encode` will escape slashes to prevent malicious strings from breaking out of script tags in HTML. Consider a string in PHP that gets encoded to JSON: 

```php
$input = "I am very </script><script>alert( 'Evil'); //";
printf( "<script>var string = %s;</script>", wp_json_encode( $string ) );
```

This prints out:

```html
<script>var string = "I am very <\/script><script>alert( 'Evil'); \/\/";</script>
```

However, with `JSON_UNESCAPED_SLASHES` it outputs:

```
<script>var string = "I am very </script><script>alert( 'Evil'); //";</script>
```

And the browser will interpret the second script tag. Basically, `json_encode` will escape slashes by default to prevent this from happening when raw JSON is printed into HTML. It eliminates the need to also worry about `esc_html()`. But in our case here, we're storing the JSON in `post_content` and aren't displaying it. So there is no need for escaping slashes. And eliminating the slashes improves readability in, for example, a changeset revisions view of when the `post_content` is printed via `wp post get 123 --field=post_content` where a URL will appear as `http://example.com/` instead of `http:\/\/example.com\/`.

More on this SO answer: http://stackoverflow.com/a/1580667/93579

ocean90: Thanks! I was more interested in the **de**coding. Turns out that `json_decode()` supports both: https://3v4l.org/3craV

src/wp-admin/export.php
ocean90: This should probably be handled through a filter or something.

src/wp-admin/js/customize-controls.js
```diff
@@ -4014,10 +4484,42 @@
...
+
+				urlParser = document.createElement( 'a' );
+				urlParser.href = location.href;
+				urlParser.search = urlParser.search.replace( /(\?|&)changeset_uuid=[^&]+(&|$)/, '$1' );
+				urlParser.search = urlParser.search.replace( /&+$/, '' );
+				if ( '' !== newStatus ) {
+					if ( urlParser.search.length > 1 ) {
+						urlParser.search += '&';
+					}
+					urlParser.search += 'changeset_uuid=' + api.settings.changeset.uuid;
```
westonruter: This should be making use of `jQuery.param()` and `wp.customize.utils.parseQueryString`

src/wp-includes/js/customize-preview.js
```diff
@@ -81,30 +116,344 @@
...
+				preview.bind( 'scroll', function( distance ) {
+					preview.window.scrollTop( distance );
+				});
+			}
 		}
 	});
 
+	/**
+	 * Inject the changeset UUID into links in the document.
+	 *
```
westonruter: This should probably be marked private. It's only supposed to be called once and isn't intended for plugins to use.

src/wp-admin/export.php
westonruter: There's no existing filter though. This `export.php` is built a bit strange in that it excludes the `_builtin` exportable posts (`post` and `page`) because those get manually listed with their own custom UI before the loop. I suppose this line could be reverted and the following `if` could be added before the loop:

```php
<?php if ( get_post_type_object( 'customize_changeset' )->can_export ) : ?>
    <p><label><input type="radio" name="content" value="<?php echo esc_attr( get_post_type_object( 'customize_changeset' )->name ); ?>" /> <?php echo esc_html( get_post_type_object( 'customize_changeset' )->label ); ?></label></p>
<?php endif; ?>
```

But that also seems unoptimal.

src/wp-includes/theme.php
```diff
@@ -2066,24 +2066,130 @@ function check_theme_switched() {
...
+	$GLOBALS['wp_customize'] = new WP_Customize_Manager( compact( 'changeset_uuid', 'theme', 'messenger_channel' ) );
+}
+
+/**
+ * Publish a snapshot's changes.
+ *
+ * @param int     $changeset_post_id Changeset post ID.
+ * @param WP_Post $changeset_post    Changeset post object.
+ */
+function _wp_customize_publish_changeset( $changeset_post_id, $changeset_post ) {
```
westonruter: It would seem perhaps that this `publish_customize_changeset` action (`{$new_status}_{$post->post_type}`) should short-circuit if the `$old_status` is also `publish`. If there is a programmatic change to a `customize_changeset` post it would not be desirable for this to cause all of the settings to be re-saved. It is somewhat hard to detect a transition from a different post status using this hook since the `$old_status` is not supplied. One option would be to add a `save_post_customize_changeset` hook which stores an `unpublished` flag to `true` in postmeta if the status is not `publish`. Then this `_wp_customize_publish_changeset` could check if that postmeta is not present, and if it is, then short-circuit. Otherwise, if it is `true` then it can proceed with the logic in this method and then at the end set the `unpublished` postmeta to `false`.

src/wp-admin/export.php
ocean90: I'd suggest to ignore `wp-admin/export.php` for now and move this to a ticket so it can be handled after the merge.

src/wp-admin/customize.php
```diff
@@ -115,7 +133,11 @@
 		<div id="customize-header-actions" class="wp-full-overlay-header">
 			<?php
 			$save_text = $wp_customize->is_theme_active() ? __( 'Save &amp; Publish' ) : __( 'Save &amp; Activate' );
-			submit_button( $save_text, 'primary save', 'save', false );
+			$save_attrs = array();
+			if ( ! current_user_can( get_post_type_object( 'customize_changeset' )->cap->publish_posts ) ) {
+				$save_attrs['style'] = 'display: none';
```
ocean90: Is this to support a "Save as Draft" button? I wonder why it's only hidden if the current user can't publish the changes.

src/wp-admin/js/customize-controls.js
```diff
@@ -3775,22 +4230,52 @@
...
+							{ setting_invalidities: settingInvalidities }
+						] );
+						api.state( 'saving' ).set( false );
+						return deferred.promise();
 					}
 
-					query = $.extend( self.query(), {
-						nonce:  self.nonce.save
+					/*
+					 * Note that exclueCustomized is intentionally not present so that the entire
```
ocean90: Typo: "d" is missing in "exclueCustomized"

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -340,25 +395,53 @@ public function doing_ajax( $action = null ) {
...
 		}
 
 		if ( ! $message ) {
 			$message = __( 'Cheatin&#8217; uh?' );
 		}
 
+		if ( $this->messenger_channel ) {
+			header( 'Content-Type: text/html; charset=' . get_option( 'blog_charset' ) );
+			echo '<!DOCTYPE html><html>';
+			wp_print_scripts( array( 'customize-base' ) );
```
ocean90: This is throwing two notices:

> Notice: wp_localize_script was called incorrectly. Scripts and styles should not be registered or enqueued until the wp_enqueue_scripts, admin_enqueue_scripts, or login_enqueue_scripts hooks. Please see Debugging in WordPress for more information. (This message was added in version 3.3.0.) in /srv/www/wp-develop/svn/src/wp-includes/functions.php on line 4064

> Notice: wp_print_scripts was called incorrectly. Scripts and styles should not be registered or enqueued until the wp_enqueue_scripts, admin_enqueue_scripts, or login_enqueue_scripts hooks. Please see Debugging in WordPress for more information. (This message was added in version 3.3.0.) in /srv/www/wp-develop/svn/src/wp-includes/functions.php on line 4064

Can be reproduced by visiting the home URL with an invalid UUID or theme like  `http://develop.wp.dev/?customize_changeset_uuid=1&customize_theme=doesnotexists&customize_messenger_channel=preview-1`

src/wp-includes/functions.php
jonathanbardo: +1 on this being expose. UUID4 function can be very useful for plugins authors.

src/wp-includes/js/customize-base.js
```diff
@@ -637,22 +637,26 @@ window.wp = window.wp || {};
...
-		 *         {string} .channel      If provided, will send the channel with each message and only accept messages a matching channel.
-		 * @param  {object} options       Extend any instance parameter or method with this object.
+		 * @param  {object} params - Parameters to configure the messenger.
+		 *         {string} params.url - The URL to communicate with.
+		 *         {window} params.targetWindow - The window instance to communicate with. Default window.parent.
+		 *         {string} params.channel - If provided, will send the channel with each message and only accept messages a matching channel.
+		 * @param  {object} options - Extend any instance parameter or method with this object.
 		 */
 		initialize: function( params, options ) {
+			var defaultTarget;
```
jonathanbardo: @westonruter Any reasons why this var was moved here? It's immediately assigned to a value from what I can see.

src/wp-includes/js/customize-base.js
```diff
@@ -637,22 +637,26 @@ window.wp = window.wp || {};
...
-			var defaultTarget = window.parent == window ? null : window.parent;
+			defaultTarget = window.parent === window ? null : window.parent;
 
 			$.extend( this, options || {} );
 
 			this.add( 'channel', params.channel );
 			this.add( 'url', params.url || '' );
 			this.add( 'origin', this.url() ).link( this.url ).setter( function( to ) {
-				return to.replace( /([^:]+:\/\/[^\/]+).*/, '$1' );
+				var urlParser = document.createElement( 'a' );
```
jonathanbardo: @westonruter That's pretty clever 😸

src/wp-includes/js/customize-loader.js
```diff
@@ -132,6 +132,19 @@ window.wp = window.wp || {};
 				targetWindow: this.iframe[0].contentWindow
 			});
 
+			// Expose the changeset UUID on the parent window's URL so that the customized state can survive a refresh.
+			if ( history.replaceState ) {
+				this.messenger.bind( 'changeset-uuid', function( changesetUuid ) {
+					var urlParser = document.createElement( 'a' );
```
jonathanbardo: @westonruter This can actually be extracted as a function if used more than once.

src/wp-includes/js/customize-preview-nav-menus.js
```diff
@@ -123,9 +123,23 @@ wp.customize.navMenusPreview = wp.customize.MenusCustomizerPreview = ( function(
...
-					delete oldValue.type_label;
-					if ( _.isEqual( oldValue, newValue ) ) {
+					_newValue = _.clone( newValue );
+					_oldValue = _.clone( oldValue );
+					delete _newValue.type_label;
+					delete _oldValue.type_label;
+
+					// Normalize URL scheme when parent frame is HTTPS to prevent selective refresh upon initial page load.
+					if ( 'https' === api.preview.scheme.get() ) {
+						urlParser = document.createElement( 'a' );
```
jonathanbardo: @westonruter Again this might be easier to read by calling a function wrapper.
Maybe something like:
```js
urlParser = api.utils.urlParser( _newValue.url );
```

src/wp-includes/functions.php
westonruter: ✅ Resolved. We'll keep `wp_generate_uuid4()`.

src/wp-admin/js/customize-controls.js
```diff
@@ -3209,72 +3403,58 @@
...
 				active: $.Deferred()
 			};
+			previewer.pendingChangesetUpdateRequests = [];
+
+			// Debounce to prevent hammering server and then wait for any pending update requests.
+			previewer.refresh = _.debounce(
+				( function( originalRefresh ) {
+					return function() {
+						var refreshOnceChangesetRequestsComplete = function() {
+							$.when.apply( $, previewer.pendingChangesetUpdateRequests ).then( function() {
```
westonruter: We could potentially re-use `api.state( 'processing' )` here instead of looking at all of the individual promises in `pendingChangesetUpdateRequests`. I suppose the only benefit of using the latter is that the refresh won't happen if one of the pending requests fail.

src/wp-includes/class-wp-customize-manager.php
westonruter: @ocean90 interesting. I see it is because of:

```php
function _wp_scripts_maybe_doing_it_wrong( $function ) {
	if ( did_action( 'init' ) ) {
		return;
	}
```

Since we're dying at the `setup_theme` action, the `init` action hasn't been called yet. Should  the condition for `did_action( 'init' )` not be replaced with:

```php
if ( did_action( 'admin_enqueue_scripts' ) || did_action( 'wp_enqueue_scripts' ) || did_action( 'login_enqueue_scripts' ) ) {
```
❓

src/wp-admin/customize.php
westonruter: That's right. If the user can't publish changes, then the button is hidden. This would only be the case if a plugin remaps the `publish_posts` meta capability to something other than `customize` that the user doesn't have. In that case, the user shouldn't see the button as it would just error each time it is pressed. So that is why it is hidden. The button remains in the document for the sake of plugins that use the `#save` selector to insert other buttons adjacent to it.

src/wp-includes/js/customize-base.js
westonruter: Good point. I believe it was because there was another variable declared there previously but which is now removed. I'll restore the original.

src/wp-includes/class-wp-customize-manager.php
westonruter: Naturally I'd add a `wp_enqueue_scripts()` call right before `wp_print_scripts( array( 'customize-base' ) )`.

westonruter:
@ocean90 @aaroncampbell  I found a very serious performance problem with how this is implemented. See details in #170 and fix in #171. The short answer is that we need to go back to using `POST` requests when doing preview refreshes, _but_ we can still use the natural URL for the iframe by doing a form submission to the previewed URL with the iframe as the target window, instead of doing an Ajax request and a `document.write()` into an `about:blank` iframe. The proposed fix includes an autosave for the changeset data at a 10 second interval, instead of updating the changeset every 250ms after any change is made. The `POST` request still references the changeset UUID and the included `customized` data is only what has been changed since the last successful changeset save.

src/wp-includes/class-wp-customize-manager.php
westonruter: @ocean90 I've fixed this via https://github.com/xwp/wordpress-develop/pull/161/commits/86eec4bc59cbb1905e01600584e1afcdf6b9da07

src/wp-includes/js/customize-preview.js
```diff
@@ -81,30 +176,398 @@
...
+	 * @since 4.7.0
+	 * @access protected
+	 *
+	 * @return {void}
+	 */
+	api.addRequestPreviewing = function addRequestPreviewing() {
+
+		/**
+		 * Rewrite Ajax requests to inject customizer state.
```
westonruter: This will not work 100% of the time, such as if an Admin Ajax handler is specifically looking for a `$_GET` param vs a `$_POST` param. But then again, the front-end shouldn't be making admin-ajax calls. If more compatibility for admin-ajax is required, then this code from Customize Snapshots can be incorporated: https://github.com/xwp/wp-customize-snapshots/blob/20051896d9274502d8101f3682f0e83d09136dbc/php/class-customize-snapshot-manager.php#L186-L249

westonruter:
Committed to core: https://core.trac.wordpress.org/changeset/38810

PR #179, opened 2016-10-19
Trac 27403: Improve UI for linking areas of Customizer preview to corresponding controls (desktop and mobile)

westonruter:
https://core.trac.wordpress.org/ticket/27403


src/wp-admin/js/customize-controls.js
```diff
@@ -5091,6 +5091,7 @@
 
 		$( '.collapse-sidebar' ).on( 'click', function() {
 			api.state( 'paneVisible' ).set( ! api.state( 'paneVisible' ).get() );
+			api.previewer.send( 'collapse-sidebar', ! api.state( 'paneVisible' ).get() );
```
westonruter: @sirbrillig Do you think we still need this now and below where this `collapse-sidebar` message toggles the body class? Users may indeed want access to the edit links when the sidebar is collapsed. And with [#36678](https://core.trac.wordpress.org/ticket/36678) it was specifically designed to cause the sidebar to expand when shift-clicking on an editable element. And so here too, when clicking on a button it should cause the sidebar to expand.

celloexpressions: And actually, this would be problematic on small screens which (I believe still) use the expanded/collapsed state. So probably okay to remove it.

src/wp-includes/js/customize-selective-refresh.js
```diff
@@ -18,6 +18,15 @@ wp.customize.selectiveRefresh = ( function( $, api ) {
...
 
+	// Extend jQuery to notice element removal and call the 'destroyed' event
+	$.event.special.destroyed = {
+		remove: function( o ) {
+			if ( o.handler ) {
+				o.handler();
+			}
+		}
+	};
+
```
sirbrillig: So that when an element (eg: widget) is removed, there's a jquery event we can respond to to remove the associated icon. Of course, now that the icon is inside the element, maybe this is no longer needed!

src/wp-includes/js/customize-selective-refresh.js
```diff
@@ -84,6 +93,7 @@ wp.customize.selectiveRefresh = ( function( $, api ) {
 			var partial = this;
 			_.each( _.pluck( partial.placements(), 'container' ), function( container ) {
 				$( container ).attr( 'title', self.data.l10n.shiftClickToEdit );
```
westonruter: Should we eliminate shift-click to edit entirely now and the tooltip that gets added for it?

sirbrillig:
FYI putting the icon inside the element can lead to artifacts like this (twentyfifteen uses `opacity: 0.7` on `.site-description` and it carries over into all child elements):



I couldn't figure out a way to override that.

celloexpressions:
There's nothing we can do about `opacity`... but hopefully that's more of an edge case. The answer for the theme is to use `rgba` instead where they can.

sirbrillig:
The following were added:

- insert icons within partials where feasible
- create a state for whether the icons are shown and hidden, and before the state is initially set, animate to show them, then set to hidden
- add a `button` with `class="screen-reader-text"` (becomes visible on focus) at the beginning of the `body` in the preview that toggles the icon visibility state.

src/wp-includes/js/customize-selective-refresh.js
```diff
@@ -99,6 +101,117 @@ wp.customize.selectiveRefresh = ( function( $, api ) {
 		},
 
 		/**
+		 * Create and show the edit icon for this element.
+		 */
+		createEditIconForElement: function( element ) {
+			var partial = this;
+			var $icon = this.createEditIcon();
+			var $element = $( element );
+			$element.prepend( $icon );
```
westonruter: @sirbrillig I'm seeing the edit icons get lost when a selective refresh is done for widgets and nav menus. This, along with the opacity issue you noted, makes me think that the edit icon should perhaps go back to being positioned outside?

westonruter:
@sirbrillig I'm not seeing the edit shortcuts for the site title or tagline anymore.

celloexpressions:
@westonruter are the partials still being added by the theme? Note that Twenty Seventeen doesn't have them yet.

src/wp-includes/js/customize-selective-refresh.js
celloexpressions: It's virtually impossible to position them with CSS (even with custom theme code) if they're outside the container and the theme, for example, floats the widgets. Let's figure out why icons aren't being re-added after a partial refreshes, then educate developers about the problem with using opacity.

sirbrillig:
>  I'm not seeing the edit shortcuts for the site title or tagline anymore.

@westonruter I made sure to refresh the icons when the partial is updated (I think the partial reference in my `partial-content-rendered` handler closure may be incorrect, but I went directly to the source for now) in b64cd1c but I'm not sure if that's the issue you're seeing. Do you mean you're not seeing them appear at all? If so, which theme are you using?

src/wp-admin/js/customize-controls.js
```diff
@@ -4154,6 +4155,15 @@
 				previewer.previewUrl.bind( previewer.refresh );
 			}
 
+			// Add invisible button to toggle editShortcutVisibility (icons)
+			var $editShortcutVisibilityButton = $( '<button type="button" class="edit-shortcut-visibility-button"><span class="screen-reader-text"></span></button>' );
```
sirbrillig: Got it. I removed the span. Tabbing to the button and clicking it (or clicking it using `document.querySelector('.edit-shortcut-visibility-button').click()` in the console) seems to work well!

westonruter:
@sirbrillig I'm using Twenty Sixteen. I'm seeing this:

customize__wordpress_develop_ _a_customizer-driven_site

sirbrillig:
Thanks! Found it. One of the containers of the header has `overflow: hidden`. Since the icon is inside the container (despite having `position: absolute` on the `::before` element itself) it is hidden when it moves outside the container.

screen shot 2016-10-25 at 3 09 44 pm

westonruter:
The build keeps failing due to JSHint issues: https://travis-ci.org/xwp/wordpress-develop/jobs/170618650#L287

src/wp-admin/js/customize-controls.js
```diff
@@ -4311,6 +4312,15 @@
 				previewer.previewUrl.bind( previewer.refresh );
 			}
 
+			// Add invisible button to toggle editShortcutVisibility (icons)
+			$editShortcutVisibilityButton = $( '<button type="button" class="edit-shortcut-visibility-button screen-reader-text"></button>' );
+			btnText = api.l10n.editShortcutIcons || 'Toggle edit icons';
+			$editShortcutVisibilityButton.find( '.screen-reader-text' ).text( btnText );
```
westonruter: This selector is failing because it _is_ the `.screen-reader-text`. It is not a parent of it.

Also, @celloexpressions didn't you intend for this button to be inside the preview and not in the controls?

src/wp-includes/js/customize-selective-refresh.js
```diff
@@ -99,6 +101,119 @@ wp.customize.selectiveRefresh = ( function( $, api ) {
...
+			}
+			if ( settingId.match( /^widget_/ ) ) {
+				return 'widget';
+			}
+			return settingId;
+		},
+
+		createEditIcon: function() {
+			var iconTitle = this.getIconTitle();
+			return $( '<button aria-label="' + iconTitle + '" title="' + iconTitle + '" type="button" class="customize-partial-icon ' + this.getIconClassName() + '"/>' );
```
westonruter: Better to let jQuery handle the parsing here to avoid HTML attribute injections. So this should be:

```js
return $( '<button>', {
    'aria-label': iconTitle,
    'title': iconTitle,
    'type': 'button',
    'className': 'customize-partial-icon ' + this.getIconClassName()
} );
```

westonruter:
* [ ] Missing JSDoc for new methods.

src/wp-includes/css/customize-preview.css
```diff
@@ -10,3 +10,153 @@
...
+	80% {
+		transform: scale3d(.97, .97, .97);
+	}
+	to {
+		opacity: 1;
+		transform: scale3d(1, 1, 1);
+	}
+}
+
+@keyframes bounce-disappear {
```
westonruter: Shouldn't this be prefixed? Be something like `customize-partial-icon-bounce-disappear`

src/wp-includes/js/customize-selective-refresh.js
```diff
@@ -99,6 +101,119 @@ wp.customize.selectiveRefresh = ( function( $, api ) {
...
+			}
+		},
+
+		getType: function() {
+			var partial = this, settingId;
+			settingId = partial.params.primarySetting || _.first( partial.settings() ) || 'unknown';
+			if ( partial.params.type ) {
+				return partial.params.type;
+			}
+			if ( settingId.match( /^nav_menu_instance/ ) ) {
```
westonruter: Safer to be `/^nav_menu_instance\[/`

src/wp-includes/js/customize-selective-refresh.js
```diff
@@ -99,6 +101,119 @@ wp.customize.selectiveRefresh = ( function( $, api ) {
...
+				return 'nav_menu';
+			}
+			if ( settingId.match( /^widget_/ ) ) {
```
westonruter: Safer to be `/^widget_.+\[\d+\]$/`

src/wp-includes/js/customize-selective-refresh.js
```diff
@@ -854,7 +976,21 @@ wp.customize.selectiveRefresh = ( function( $, api ) {
 			}
 		} );
 
+		api.preview.bind( 'edit-shortcut-visibility', function( visibility ) {
+			api.selectiveRefresh.editShortcutVisibility.set( visibility );
+		} );
+		api.selectiveRefresh.editShortcutVisibility.bind( function( visibility ) {
+			var body = $( document.body ), shouldAnimateHide;
+
+			// @todo Handle initial state.
```
sirbrillig: Oh! yes, sorry.

src/wp-includes/js/customize-selective-refresh.js
```diff
@@ -99,6 +101,205 @@ wp.customize.selectiveRefresh = ( function( $, api ) {
...
+		 * @returns {void}
+		 */
+		positionEditShortcut: function( placement, $editShortcut ) {
+			var partial = this, $placementContainer;
+			$placementContainer = $( placement.container );
+			$placementContainer.prepend( $editShortcut );
+			if ( 'absolute' === $placementContainer.css( 'position' ) ) {
+				$editShortcut.addClass( 'customize-partial-edit-shortcut-absolute' );
+				$editShortcut.css( partial.getEditShortcutPositionStyles( $placementContainer ) );
+				partial.whenPageChanges( function() {
```
westonruter: What about when the partial is removed from the document? This “event” listener should get unbound.

src/wp-includes/js/customize-selective-refresh.js
```diff
@@ -99,6 +101,205 @@ wp.customize.selectiveRefresh = ( function( $, api ) {
...
+			partial.positionEditShortcut( placement, $shortcut );
+			$shortcut.on( 'click', function( event ) {
+				event.preventDefault();
+				event.stopPropagation();
+				partial.showControl();
+			} );
+			api.selectiveRefresh.bind( 'partial-content-rendered', function( placement ) {
+				partial.refreshEditShortcut( placement );
+			} );
+			api.selectiveRefresh.bind( 'partial-content-moved', function( placement ) {
```
westonruter: When the partial is removed, shouldn't these event listeners get removed as well?

src/wp-includes/js/customize-selective-refresh.js
sirbrillig: Oo, yup. This does have a memory leak.

sirbrillig:
9ba5bfe prevents the memory leak by disabling the listeners when the element no longer exists, but what I'd rather do is remove the listeners when the partial is removed. I just can't find the place in the code where that happens.

westonruter:
@sirbrillig you can listen for partials being removed via:

```js
wp.customize.selectiveRefresh.partial.bind( 'remove', function () {
    // Remove event listeners, including this one.
} );
```

sirbrillig:
Thanks, @westonruter! I couldn't actually get that working, but also it appears that the absolute positioning code is no longer required. I'm... not exactly sure why it was needed previously and not any more, but in the themes I've tested I can't find a place where the default button positioning doesn't work. Maybe one of the CSS changes that Nick made took care of the issue I had been handling?

In any case, for now I've removed it all in e61aac5, which is a lot of code saved.

westonruter:
Committed to core in:
https://core.trac.wordpress.org/changeset/38967
https://core.trac.wordpress.org/changeset/38972

PR #180, opened 2016-10-21
Add Notifications Area in Customizer (#35210, #37727)

delawski:
https://core.trac.wordpress.org/ticket/35210


src/wp-includes/class-wp-customize-manager.php
```diff
@@ -2133,6 +2134,29 @@ public function refresh_nonces() {
 	}
 
 	/**
+	 * Render JS templates for all common elements.
+	 *
+	 * @since 4.7.0
+	 * @access public
+	 */
+	public function render_common_templates() {
```
delawski: @westonruter I was thinking of using one common template for both _global_ and _control-specific_ notifications. Right now I'm using this template only in _global_ notification area, but I could possibly use instead of this one for control specific notifications:

https://github.com/xwp/wordpress-develop/blob/master/src/wp-includes/class-wp-customize-manager.php#L2494-L2500

In terms of the markup, there would be no difference to the control notification ('dismiss' button has to be explicitly turned on and UL class is optional).

What do you think?

src/wp-includes/js/customize-base.js
```diff
@@ -791,6 +791,123 @@ window.wp = window.wp || {};
...
+		 * @since 4.7.0
+		 * @param {Event} e
+		 * @this {wp.customize.notifications}
+		 */
+		dismissNotification: function( e ) {
+			var self = this,
+				code = $( e.currentTarget ).closest( '[data-code]' ).data( 'code' );
+
+			if ( code && self.has( code ) ) {
+				self.remove( code );
```
delawski: @westonruter Can you take a look at this? I've debugged the code and even though I'm removing a notification here, it still exists in the collection in the `render()` method (where `render` is bound to the collection's `remove` event in `initialize()` method). Am I missing something so that the 'removal' action gets actually 'applied' on the collection?

westonruter: @delawski I believe the problem is that the `remove` event is triggered immediately _before_ the item is removed from the collection. In the notifications collection manager in the control, you'll see that it uses a `debounce` to get around this, so that when the collection gets re-rendered the item will have indeed been removed. See also the proposal to add a `removed` event in https://core.trac.wordpress.org/ticket/37269

src/wp-includes/class-wp-customize-manager.php
westonruter: Having a common template makes sense. So you're saying the `button` would only be rendered if the `Notification` has the `isDismissible` property set, right? There would definitely be cases where a notification on a control should also be dismissible. So there should be event handlers for these buttons in control notifications in addition to the global notifications. Good on you for using a property that mirrors the `is-dismissible` class for admin notices.

src/wp-includes/js/customize-base.js
```diff
@@ -808,6 +808,170 @@ window.wp = window.wp || {};
...
+		 * @type {string}
+		 */
+		parentSelector: '#widgets-right',
+
+		/**
+		 * Initialize notifications area.
+		 *
+		 * @since 4.9.0
+		 * @constructor
+		 * @param {object} options - Options.
```
westonruter: Todo: There should be a `container` arg that is passed here.

src/wp-includes/js/customize-base.js
```diff
@@ -808,6 +808,170 @@ window.wp = window.wp || {};
...
+		 */
+		defaultTemplateName: 'customize-control-notifications',
+
+		/**
+		 * The notifications area parent container selector.
+		 *
+		 * @since 4.9.0
+		 * @type {string}
+		 */
+		parentSelector: '#widgets-right',
```
westonruter: Todo: This can be removed in favor of a `container` arg being supplied.

src/wp-includes/js/customize-base.js
```diff
@@ -808,6 +808,170 @@ window.wp = window.wp || {};
...
+		},
+
+		/**
+		 * Find and return notifications area container element.
+		 *
+		 * @since 4.9.0
+		 * @returns {jQuery} Container jQuery element or an empty DIV.
+		 * @this {wp.customize.notifications}
+		 */
+		getContainer: function() {
```
westonruter: Todo: This can be removed in favor of just getting the `container` property.

src/wp-includes/js/customize-base.js
```diff
@@ -808,6 +808,170 @@ window.wp = window.wp || {};
...
+				if ( ! parent || ! parent.length ) {
+					parent = $( '<div></div>' );
+				}
+				self.container = $( '<div></div>', {
+					id: 'customize-notifications-area',
+					'class': 'customize-control-notifications-container'
+				} );
+				parent.prepend( self.container );
+
+				self.container.on( 'click', '.notice-dismiss', self.dismissNotification );
```
westonruter: Todo: This can be added to the `initialize` method.

src/wp-includes/js/customize-base.js
```diff
@@ -808,6 +808,170 @@ window.wp = window.wp || {};
...
+				parent = $( self.parentSelector );
+
+				if ( ! parent || ! parent.length ) {
+					parent = $( '<div></div>' );
+				}
+				self.container = $( '<div></div>', {
+					id: 'customize-notifications-area',
+					'class': 'customize-control-notifications-container'
+				} );
+				parent.prepend( self.container );
```
westonruter: By passing the `container` into the constructed `Notifications` we can eliminate the need for this code. It makes more sense to me to not hard-code selectors in this collection.

src/wp-includes/js/customize-base.js
```diff
@@ -808,6 +808,170 @@ window.wp = window.wp || {};
...
+		},
+
+		/**
+		 * Remove notifications area from the DOM.
+		 *
+		 * @since 4.9.0
+		 * @returns {void}
+		 * @this {wp.customize.notifications}
+		 */
+		destroy: function() {
```
westonruter: This can be removed. The container should never get destroyed. It should only get hidden if there are no notifications remaining.

src/wp-includes/js/customize-base.js
```diff
@@ -808,6 +808,170 @@ window.wp = window.wp || {};
...
+			} else {
+				container.empty().append( $.trim(
+					template( { notifications: notifications } )
+				) );
+				sidebar.css( 'top', '' );
+				containerHeight = container.outerHeight() + 1;
+				containerInitialTop = parseInt( sidebar.css( 'top' ), 10 );
+				sidebar
+					.css( 'top', containerInitialTop + containerHeight + 'px' )
+					.trigger( 'customize:sidebar:updateLayout' );
```
westonruter: This can be replaced with `self.trigger( 'rendered' )`. Then the rest of the Customizer can watch for the event via `notifications.bind( 'rendered', () => { … } )`

src/wp-includes/js/customize-base.js
```diff
@@ -808,6 +808,170 @@ window.wp = window.wp || {};
...
+				self.destroy();
+			} else {
+				container.empty().append( $.trim(
+					template( { notifications: notifications } )
+				) );
+				sidebar.css( 'top', '' );
+				containerHeight = container.outerHeight() + 1;
+				containerInitialTop = parseInt( sidebar.css( 'top' ), 10 );
+				sidebar
+					.css( 'top', containerInitialTop + containerHeight + 'px' )
```
westonruter: Since the above logic is specific to the global notifications are, it might be better to be put in an overridden `render` method when initialized for `wp.customize.notifications`. Or rather, the logic can be added into a `notifications.bind( 'rendered' )` event handler.

src/wp-includes/js/customize-base.js
```diff
@@ -808,6 +808,170 @@ window.wp = window.wp || {};
...
+		 * @since 4.9.0
+		 * @param {Event} event - Event.
+		 * @returns {void}
+		 * @this {wp.customize.notifications}
+		 */
+		dismissNotification: function( event ) {
+			var self = this,
+				code = $( event.currentTarget ).closest( '[data-code]' ).data( 'code' );
+
+			if ( code && self.has( code ) ) {
```
westonruter: This should also do `self.trigger( 'dismissed', self( code ) )` so that we can handle an action for dismissal. Also, we should be able to do `self( code ).trigger( 'dismissed' )` so that the `Notification` itself can be aware of when it was dismissed.

westonruter:
- [ ] We should consider adding the `slideUp` animation when dismissing or removing a notification.
- [ ] The `api.Notifications` collection should get utilized for controls as well.

tests/qunit/index.html
```diff
@@ -203,6 +203,18 @@ <h3 class="accordion-section-title" tabindex="0">
...
+				<# _.each( data.notifications, function( notification ) { #>
+					<li data-code="{{ notification.code }}" data-type="{{ notification.type }}">
+						<p>{{ notification.message || notification.code }}</p>
+						<# if ( notification.isDismissible ) { #>
+							<button type="button" class="notice-dismiss"><span class="screen-reader-text">Dismiss</span></button>
+						<# } #>
+					</li>
+				<# } ); #>
+			</ul>
+		</script>
```
westonruter: I forgot to remove this.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -3183,7 +3183,12 @@ public function render_control_templates() {
 		<script type="text/html" id="tmpl-customize-control-notifications">
 			<ul>
 				<# _.each( data.notifications, function( notification ) { #>
-					<li class="notice notice-{{ notification.type || 'info' }} {{ data.altNotice ? 'notice-alt' : '' }}" data-code="{{ notification.code }}" data-type="{{ notification.type }}">{{{ notification.message || notification.code }}}</li>
+					<li class="notice notice-{{ notification.type || 'info' }} {{ data.altNotice ? 'notice-alt' : '' }} {{ notification.isDismissible ? 'is-dismissible' : '' }}" data-code="{{ notification.code }}" data-type="{{ notification.type }}">
+						{{{ notification.message || notification.code }}}
+						<# if ( notification.isDismissible ) { #>
```
westonruter: We might as well extract this content here to be put in a separate `tmpl-customize-notification` (singular). A notification then can itself have a `render` method which would by default just apply its data to a `tmpl-customize-notification` along with the `data` supplied here. This then would resolve [#37727](https://core.trac.wordpress.org/ticket/37727).

westonruter: Maybe `notification.render()` is wrong as well. The notification can just render itself when it is constructed and then it can have a `notification.container` element that has the rendered element, which then can have it's own custom elements inside of it and event handlers added to them. It can even handle dismissing itself by calling `remove` on its `parent` Notifications object.

In this case, we could eliminate `tmpl-customize-control-notifications` entirely, and the logic for updating notifications can just be to make sure that all of the notifications' `container` elements are added to the appropriate location.

westonruter:
Committed to trunk in [r41374](https://core.trac.wordpress.org/changeset/41374).

PR #182, opened 2016-10-25
Trac 38164: Customize: assign static front page and posts page to new pages

westonruter:
https://core.trac.wordpress.org/ticket/38164

### DIFF
--- a/src/wp-admin/js/customize-controls.js
+++ b/src/wp-admin/js/customize-controls.js
@@ -2762,12 +2762,12 @@
 				return;
 			}
 
-			var control = this, promise,
-			    toggle = control.container.find( '.add-new-toggle' ),
-			    container = control.container.find( '.new-content-item' ),
-			    input = control.container.find( '.create-item-input' ),
-			    title = input.val();
-			    select = control.container.find( 'select' );
+			var control = this, promise, toggle, container, input, title, select;
+			toggle = control.container.find( '.add-new-toggle' );
+			container = control.container.find( '.new-content-item' );
+			input = control.container.find( '.create-item-input' );
+			title = input.val();
+			select = control.container.find( 'select' );
 
 			if ( ! title ) {
 				input.addClass( 'invalid' );
--- a/src/wp-admin/js/customize-controls.js
+++ b/src/wp-admin/js/customize-controls.js
@@ -2808,7 +2808,6 @@
 				// Focus the select control.
 				select.focus();
 				control.setting.set( String( data.post_id ) ); // Triggers a preview refresh and updates the setting.
-				select.val( String( data.post_id ) );
 
 				// Reset the create page form.
 				container.slideUp( 180 );
--- a/src/wp-admin/js/customize-controls.js
+++ b/src/wp-admin/js/customize-controls.js
@@ -2512,12 +2512,12 @@
 		/**
 		 * Triggered when the control's markup has been injected into the DOM.
 		 *
-		 * @abstract
+		 * @returns {void}
 		 */
 		ready: function() {
-			if ( 'dropdown-pages' === this.params.type && this.params.allow_addition ) {
-				var control = this,
-				    newItem = control.container.find( '.new-content-item' );
+			var control = this, newItem;
+			if ( 'dropdown-pages' === control.params.type && control.params.allow_addition ) {
+				newItem = control.container.find( '.new-content-item' );
 				newItem.hide(); // Hide in JS to preserve flex display when showing.
 				control.container.on( 'click', '.add-new-toggle', function( e ) {
 					$( e.currentTarget ).slideUp( 180 );
@@ -2757,11 +2757,12 @@
 		 * Add a new page to a dropdown-pages control reusing menus code for this.
 		 *
 		 * @since 4.7.0
+		 * @returns {void}
 		 */
 		addNewPage: function () {
 			var control = this, promise, toggle, container, input, title, select;
 
-			if ( 'dropdown-pages' !== this.params.type || ! this.params.allow_addition || ! api.Menus ) {
+			if ( 'dropdown-pages' !== control.params.type || ! control.params.allow_addition || ! api.Menus ) {
 				return;
 			}
 
--- a/src/wp-admin/js/customize-controls.js
+++ b/src/wp-admin/js/customize-controls.js
@@ -2514,7 +2514,25 @@
 		 *
 		 * @abstract
 		 */
-		ready: function() {},
+		ready: function() {
+			if ( 'dropdown-pages' === this.params.type ) {
+				var control = this;
+				control.container.find( '.new-content-item' ).hide(); // Hide in JS to preserve flex display when showing.
+				control.container.on( 'click', '.add-new-toggle', function( e ) {
+					$( e.currentTarget ).slideUp( 180 );
+					control.container.find( '.new-content-item' ).slideDown( 180 )
+					                                             .find( '.create-item-input' ).focus();
+				});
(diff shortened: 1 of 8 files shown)


src/wp-includes/class-wp-customize-control.php
```diff
@@ -557,7 +563,14 @@ protected function render_content() {
 				echo $dropdown;
 				?>
 				</label>
-				<?php
+				<?php if ( $this->add_new && current_user_can( 'publish_pages' ) && current_user_can( 'edit_theme_options' ) ) : // Currently tied to menus functionality. ?>
+					<button type="button" class="button add-new-toggle"><?php _e( 'Add new page' ); ?></button>
```
westonruter: This should re-use `get_post_type_object( 'page' )->labels->add_new_item`

westonruter: Also, this button is not appearing on a new line for me.

src/wp-includes/class-wp-customize-nav-menus.php
```diff
@@ -370,6 +370,7 @@ public function enqueue_scripts() {
 				'untitled'          => _x( '(no label)', 'missing menu item navigation label' ),
 				'unnamed'           => _x( '(unnamed)', 'Missing menu name.' ),
 				'custom_label'      => __( 'Custom Link' ),
+				'page_label'        => __( 'Page' ),
```
westonruter: This can re-use `get_post_type_object( 'page' )->labels->singular_name`.

src/wp-admin/js/customize-controls.js
```diff
@@ -2732,6 +2750,71 @@
...
+		 * Add a new page to a dropdown-pages control reusing menus code for this.
+		 *
+		 * @since 4.7.0
+		 */
+		addNewPage: function () {
+			if ( 'dropdown-pages' !== this.params.type || ! api.Menus ) {
+				return;
+			}
+
+			var control = this, promise, toggle, container, input, title, select;
```
westonruter: The `var` should be the first statement in a function.

src/wp-admin/js/customize-controls.js
```diff
@@ -2732,6 +2750,71 @@
...
+				$content.prepend( itemTemplate( availableItem.attributes ) );
+
+				// Focus the select control.
+				select.focus();
+				control.setting.set( String( data.post_id ) ); // Triggers a preview refresh and updates the setting.
+				select.val( String( data.post_id ) );
+
+				// Reset the create page form.
+				container.slideUp( 180 );
+				toggle.slideDown( 180 );
```
westonruter: Could/should CSS transitions be used here instead? /cc @delawski

src/wp-admin/js/customize-controls.js
```diff
@@ -2732,6 +2750,71 @@
...
+				// Add the new item to the list of available menu items.
+				api.Menus.availableMenuItemsPanel.collection.add( availableItem );
+				$content = $( '#available-menu-items-post_type-page' ).find( '.available-menu-items-list' );
+				itemTemplate = wp.template( 'available-menu-item' );
+				$content.prepend( itemTemplate( availableItem.attributes ) );
+
+				// Focus the select control.
+				select.focus();
+				control.setting.set( String( data.post_id ) ); // Triggers a preview refresh and updates the setting.
+				select.val( String( data.post_id ) );
```
westonruter: This is not working for me. The `option` is getting added by `insertAutoDraftPost` when the promise is `done` there. There may be a race condition. Probably safest to append the `option` in here if it isn't already appended, or modify `insertAutoDraftPost` to return a new promise that resolves after the `request.done()` finishes in `insertAutoDraftPost`.

westonruter: Also, this line shouldn't be necessary at all. Merely calling `control.setting.set( String( data.post_id ) )` will cause the selected option to change because of the `Element` two-way data binding between the `select` and the `setting`.

src/wp-admin/js/customize-controls.js
```diff
@@ -2514,7 +2514,25 @@
...
 		 */
-		ready: function() {},
+		ready: function() {
+			if ( 'dropdown-pages' === this.params.type ) {
+				var control = this;
+				control.container.find( '.new-content-item' ).hide(); // Hide in JS to preserve flex display when showing.
+				control.container.on( 'click', '.add-new-toggle', function( e ) {
+					$( e.currentTarget ).slideUp( 180 );
+					control.container.find( '.new-content-item' ).slideDown( 180 )
+					                                             .find( '.create-item-input' ).focus();
```
westonruter: Let's change this to:

```js
control.container.find( '.new-content-item' )
	.slideDown( 180 )
	.find( '.create-item-input' ).focus()
```

Or rather:

```js
var newContentItem = control.container.find( '.new-content-item' );
newContentItem.slideDown( 180 );
newContentItem.find( '.create-item-input' ).focus();
```

src/wp-includes/class-wp-customize-control.php
```diff
@@ -115,6 +115,12 @@ class WP_Customize_Control {
 	public $input_attrs = array();
 
 	/**
+	 * @access public
+	 * @var array
+	 */
+	public $add_new = false;
```
westonruter: Maybe `allow_addition` or `addition_ui` or something like that would be clearer?

src/wp-admin/js/customize-controls.js
```diff
@@ -2514,7 +2514,25 @@
 		 *
 		 * @abstract
 		 */
-		ready: function() {},
+		ready: function() {
+			if ( 'dropdown-pages' === this.params.type ) {
```
westonruter: This should also look for `this.params.add_new`, which should be included in the `json` exports I should think.

src/wp-includes/class-wp-customize-control.php
westonruter: Also, consider that this is being added to the _base_ `WP_Customize_Control` that everything else inherits from. So in lieu of there being a separate `WP_Customize_Dropdown_Pages_Control` maybe there should be a prefixed property name like `dropdown_pages_addition_ui`.

src/wp-admin/js/customize-controls.js
celloexpressions: We could, but slideUp/Down is more elegant here since we'd have to toggle a container class anyway, and also add CSS to show/hide and animate. We use the jQuery slide functions in a few places in the customizer where it makes sense for secondary elements (there's a couple in the new themes panel, for example).

src/wp-includes/class-wp-customize-control.php
celloexpressions: I'm not sure that there are any advantages to prefixing it; it probably makes more sense without that aspect in terms of adding support for it when you register custom dropdown-pages controls. It should be fairly obvious that it doesn't apply to every type, and I'll indicate that in a comment as well.

src/wp-admin/js/customize-controls.js
westonruter: Yes, works well now! 👍

src/wp-includes/class-wp-customize-control.php
```diff
@@ -296,6 +304,10 @@ public function to_json() {
 		$this->json['label'] = $this->label;
 		$this->json['description'] = $this->description;
 		$this->json['instanceNumber'] = $this->instance_number;
+
+		if ( 'dropdown-pages' === $this->type ) {
```
westonruter: Excellent.

westonruter:
Committed to core in https://core.trac.wordpress.org/changeset/38906

PR #183, opened 2016-10-25
Trac 38377: Empty favicons cause extra HTTP requests to customizer when history.replaceState() used

westonruter:
https://core.trac.wordpress.org/ticket/38377


westonruter:
Committed in https://core.trac.wordpress.org/changeset/38901

PR #184, opened 2016-10-26
Trac 22058: Improve custom background properties UI

westonruter:
https://core.trac.wordpress.org/ticket/22058


src/wp-admin/js/customize-controls.js
```diff
@@ -5557,6 +5590,93 @@
...
+				if ( 'custom' === preset ) {
+					return;
+				}
+
+				updateSettings( preset );
+			} );
+		} );
+
+		api.control( 'background_repeat', function( control ) {
+			control.elements[0].unsync( api( 'background_repeat' ) );
```
westonruter: This confused me.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -3306,36 +3309,65 @@ public function register_controls() {
...
+			'section'  => 'background_image',
+			'settings' => array(
+				'x' => 'background_position_x',
+				'y' => 'background_position_y',
+			),
+		) ) );
+
+		$this->add_setting( 'background_size', array(
+			'default'        => get_theme_support( 'custom-background', 'default-size' ),
+			'theme_supports' => 'custom-background',
```
westonruter: Missing `sanitize_callback`.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -3284,20 +3286,21 @@ public function register_controls() {
 
 		$this->add_control( new WP_Customize_Background_Image_Control( $this ) );
 
-		$this->add_setting( 'background_repeat', array(
-			'default'        => get_theme_support( 'custom-background', 'default-repeat' ),
+		$this->add_setting( 'background_preset', array(
+			'default'        => get_theme_support( 'custom-background', 'default-preset' ),
```
westonruter: Missing `sanitize_callback`.

src/wp-admin/js/customize-controls.js
westonruter: So the reason for this is that the a `checkbox` in core by default will only ever store a boolean value, right? There is a bit of a better approach which could be considered here and incorporated into core: https://github.com/xwp/wp-customize-posts/blob/0.8.2/js/customize-post-section.js#L12-L34

westonruter: Nevermind, what you have in place is following an existing pattern used by `display_header_text`. So it's not any worse than what is already in core.

cdog: Yes, this is for the checkbox to return 'repeat'/'no-repeat' instead of true/false.

src/wp-admin/js/customize-controls.js
```diff
@@ -5557,6 +5590,93 @@
...
+				control.setting.set( to ? 'repeat' : 'no-repeat' );
+			} );
+
+			control.setting.bind( function( to ) {
+				control.element.set( 'no-repeat' !== to );
+			} );
+		} );
+
+		api.control( 'background_attachment', function( control ) {
+			control.elements[0].unsync( api( 'background_attachment' ) );
```
cdog: Same as background_repeat, follows display_header_text already in core.

PR #186, opened 2016-10-27
Trac 38499 : Display checkmark on newly-created pages.

kienstra:
**Request For Review Here Or On [Trac Ticket](https://core.trac.wordpress.org/ticket/38499)**

Hi @celloexpressions,
If you'd like to review this pull request here, that'd be great. Otherwise, looking at the patch on the [Trac ticket](https://core.trac.wordpress.org/ticket/38499) would be just as good.

This passes the [available-menu-item template](https://core.trac.wordpress.org/ticket/38499#comment:3) a value of `true` for [is_new_item](https://github.com/xwp/wordpress-develop/compare/trac-38499?expand=1#diff-027b34627f17d21422fd5065315a6226R589). The template will then output the class `item-added`, as you [suggested](https://core.trac.wordpress.org/ticket/38499#trac-change-2-1477591722273107).

### DIFF
--- a/src/wp-admin/js/customize-nav-menus.js
+++ b/src/wp-admin/js/customize-nav-menus.js
@@ -585,7 +585,8 @@
 					'type_label': itemTypeLabel,
 					'object': itemObject,
 					'object_id': data.post_id,
-					'url': data.url
+					'url': data.url,
+					'is_new_item': true
 				} );
 
 				// Add new item to menu.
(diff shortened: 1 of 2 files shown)


celloexpressions:
Could we add the class in the JS right after inserting the available item to the top of the list, rather than adding a new property to it?

kienstra:
**An Approach With Only JavaScript**

Hi @celloexpressions,
Thanks for reviewing this PR. Is this [patch](https://core.trac.wordpress.org/attachment/ticket/38499/alt-38499.diff) what you had in mind in your [comment](https://github.com/xwp/wordpress-develop/pull/186#issuecomment-256800356)?

PR #187, opened 2016-10-27
Change edit shortcut buttons to buttons in spans

sirbrillig:
Replaces click target pseudo elements (`button:before`) with a `button` and replaces the `button` with a `span`.

Fixes https://core.trac.wordpress.org/ticket/38532

The styles are all messed up and need to be fixed before this can be merged.


sirbrillig:
@celloexpressions: I think I got the icons working. Can you try this out?

src/wp-includes/js/customize-selective-refresh.js
```diff
@@ -1024,7 +1030,7 @@ wp.customize.selectiveRefresh = ( function( $, api ) {
 			} );
 
 			body.on( 'click', function( event ) {
-				if ( $( event.target ).is( '.customize-partial-edit-shortcut, :input, a[href]' ) || 0 !== $( event.target ).closest( 'a' ).length ) {
+				if ( event.shiftKey || $( event.target ).is( '.customize-partial-edit-shortcut, :input, a[href]' ) || 0 !== $( event.target ).closest( 'a' ).length ) {
```
westonruter: @sirbrillig FYI, I included this change as part of the fix for #38554. See https://core.trac.wordpress.org/changeset/39018

sirbrillig: Thanks! Rebased.

src/wp-includes/css/customize-preview.css
```diff
@@ -28,23 +28,24 @@
...
-.customize-partial-edit-shortcut:before {
+.widget .customize-partial-edit-shortcut button,
+.customize-partial-edit-shortcut button {
 	position: absolute;
 	left: -36px;
 	color: #fff;
-	width: 30px;
-	height: 30px;
+	min-width: 30px;
+	min-height: 30px;
```
delawski: Any particular reason for having `min-width/min-height` instead of just `width/height`? It results in icons not being scaled uniformly like here (Twenty Eleven):

Non-uniformly scaled icons

Also, the icons are bigger than 30x30px on some themes (e.g. Twenty Seventeen) because of having only `min` constraints.

sirbrillig:
Updated the only theme overrides I found (twentyfourteen and twentyseventeen).

src/wp-includes/css/customize-preview.css
delawski: Yeah, it doesn't show for me too if only `width/height` is applied. But I was able to get best results when both `width/height` and `min-width/min-height` were applied at the same time.

sirbrillig: I don't see why not. Added.

sirbrillig:
Rebased to fix conflicts

PR #190, opened 2016-10-30
WP Trac 38539: Customize: Prevent auto-draft post/page stubs from being saved with empty slugs.

westonruter:
https://core.trac.wordpress.org/ticket/38539


westonruter:
Committed in https://core.trac.wordpress.org/changeset/39038

PR #193, opened 2016-11-04
#37269 - Add `removed` event to `Values` collection (with unit tests).

delawski:


src/wp-includes/js/customize-base.js
```diff
@@ -438,6 +438,7 @@ window.wp = window.wp || {};
 
 			delete this._value[ id ];
 			delete this._deferreds[ id ];
+			this.trigger( 'removed', value );
```
westonruter: Should this be done only if `! _.isUndefined( value )`? In other words, I would think that the `removed` event should only fire if the `remove` event also fired.

westonruter:
Merged into #180 as https://github.com/xwp/wordpress-develop/pull/180/commits/8bbe1f12e2a27bd143b3df3fada3e377ec92f6cf

PR #194, opened 2016-11-07
#37471 - Open section automatically if panel contains only one section.

delawski:
https://core.trac.wordpress.org/ticket/37471


westonruter:
Committed to trunk in [r40395](https://core.trac.wordpress.org/changeset/40395), minus d23b3d5.

PR #196, opened 2016-11-08
WP Trac 38705: Store user ID with setting change written into changeset and restore user when setting is saved

westonruter:
https://core.trac.wordpress.org/ticket/38705


westonruter:
Committed in r39181.

PR #198, opened 2016-11-08
WP Trac 38719: Improve trashing behavior of customize_changeset posts

westonruter:
https://core.trac.wordpress.org/ticket/38719


westonruter:
Committed in r39180.

PR #199, opened 2016-11-11
WP Trac 38672: Custom CSS should work with existing Jetpack custom CSS

westonruter:
See https://core.trac.wordpress.org/ticket/38672
Demo plugin that makes use of patch: https://github.com/xwp/wp-custom-scss-demo

 Customize: Improve extensibility of Custom CSS
* Eliminate `final`-ity from `WP_Customize_Custom_CSS_Setting` class to allow extensibility.
* Eliminate use of `wp_get_custom_css()` when getting the setting value. Use the underlying `post_value` directly when `is_previewed`.
* Make clear that `wp_get_custom_css()` is specifically for obtaining the value to render/display.
* Move anonymous functions handing JS previewing for `custom_logo`, `custom_css`, and `background` into named functions on the `wp.customize.settingPreviewHandlers` to allow plugins to override/extend.
* Update `_custom_background_cb` to always print a `style` tag in the customizer preview.
* Update `background` preview logic to replace existing `style` element instead of appending a new `style` to the `head`, and thus unexpectedly overriding any Custom CSS in the preview.

### DIFF
--- a/src/wp-includes/customize/class-wp-customize-custom-css-setting.php
+++ b/src/wp-includes/customize/class-wp-customize-custom-css-setting.php
@@ -268,12 +268,12 @@ final class WP_Customize_Custom_CSS_Setting extends WP_Customize_Setting {
 		 *         $css = $post->post_content_filtered;
 		 *     }
 		 *     return $css;
-		 * }
+		 * }, 10, 2 );
 		 * </code>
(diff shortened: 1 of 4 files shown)


westonruter:
Committed to core in r39209.

PR #202, opened 2016-11-19
#38865: Customize: Setting validation error messages can fail to clear if newly-supplied valid value matches previously-valid value stored in changeset

westonruter:
https://core.trac.wordpress.org/ticket/38865

* Customize: Ensure that `WP_Customize_Manager::save_changeset_post()` returns `setting_validities` even for supplied values that are unchanged.
* Check setting existence and authorization via `WP_Customize_Manager::validate_setting_values()` even for `null` values.


westonruter:
Committed in r39320.

PR #203, opened 2016-11-21
Trac 38867: Changesets : opening a link from the preview as a new browser tab can be confusing

westonruter:
https://core.trac.wordpress.org/ticket/38867


westonruter:
Committed

PR #204, opened 2016-11-22
#38672: Add wp_update_custom_css_post() function and tests

westonruter:
https://core.trac.wordpress.org/ticket/38672

### DIFF
--- a/src/wp-includes/customize/class-wp-customize-custom-css-setting.php
+++ b/src/wp-includes/customize/class-wp-customize-custom-css-setting.php
@@ -245,13 +245,9 @@ final class WP_Customize_Custom_CSS_Setting extends WP_Customize_Setting {
 			$css = '';
 		}
 
-		$args = array(
-			'post_content' => $css,
-			'post_content_filtered' => '',
-		);
-
-		$args['stylesheet'] = $this->stylesheet;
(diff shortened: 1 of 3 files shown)


src/wp-includes/customize/class-wp-customize-custom-css-setting.php
```diff
@@ -256,39 +256,8 @@ public function update( $css ) {
...
-		 * }
-		 * @param string                          $css     Original CSS being updated.
-		 * @param WP_Customize_Custom_CSS_Setting $setting Custom CSS Setting.
-		 */
-		$args = apply_filters( 'customize_update_custom_css_post_content_args', $args, $css, $setting );
-		$args = wp_array_slice_assoc( $args, array( 'post_content', 'post_content_filtered' ) );
-
 		$args['stylesheet'] = $this->stylesheet;
-		$r = wp_update_custom_css_post( $args );
+		$r = wp_update_custom_css_post( $css, $args );
```
georgestephanis: If we're doing it that way, I wonder if the function name should change from `wp_update_custom_css_post()` to `wp_update_custom_css()` -- as $css and possibly only argument, the caller probably doesn't care about the post.

PR #206, opened 2016-11-27
WP Trac #38955: Customize Menus: post type archive menu items lose their labels when loading the customizer

westonruter:
https://core.trac.wordpress.org/ticket/38955


westonruter:
Committed in r39393.

PR #207, opened 2016-11-27
Trac 38952: Customize Menus: screen options are broken

westonruter:
https://core.trac.wordpress.org/ticket/38952


westonruter:
Committed in r39378.

PR #208, opened 2016-11-28
Trac: 38943

PatelUtkarsh:
https://core.trac.wordpress.org/ticket/38943

### DIFF
--- a/src/wp-includes/class-wp-customize-manager.php
+++ b/src/wp-includes/class-wp-customize-manager.php
@@ -2036,20 +2036,17 @@ final class WP_Customize_Manager {
 			}
 		}
 
-		$changeset_post = get_post( $changeset_post_id );
 		$now = gmdate( 'Y-m-d H:i:59' );
-		$is_valid_changeset_post_and_future_date = ( $changeset_post instanceof WP_Post
-		                                             &&
-		                                             ( mysql2date( 'U', $changeset_post->post_date_gmt, false ) > mysql2date( 'U', $now, false ) ) );
+		$is_updating_future_dated_changeset = false;
+		if ( $changeset_post_id ) {
+			$changeset_post = get_post( $changeset_post_id );
+			$is_updating_future_dated_changeset = mysql2date( 'U', $changeset_post->post_date_gmt, false ) > mysql2date( 'U', $now, false );
+		}
 
 		if ( isset( $changeset_date_gmt ) ) {
 			$is_future_dated = ( mysql2date( 'U', $changeset_date_gmt, false ) > mysql2date( 'U', $now, false ) );
 			if ( ! $is_future_dated ) {
-				if ( 'future' === $changeset_status ) {
-					wp_send_json_error( 'not_future_date', 400 ); // Only future dates are allowed.
-				} else {
-					$changeset_date_gmt = '0000-00-00 00:00:00';
-				}
+				wp_send_json_error( 'not_future_date', 400 ); // Only future dates are allowed.
 			}
 
 			if ( ! $this->is_theme_active() && $is_future_dated ) {
@@ -2059,16 +2056,9 @@ final class WP_Customize_Manager {
 			if ( $will_remain_auto_draft ) {
 				wp_send_json_error( 'cannot_supply_date_for_auto_draft_changeset', 400 );
 			}
-		} elseif ( ! $is_valid_changeset_post_and_future_date && 'future' === $changeset_status ) {
-			// Bail if date is not passed existing date is not in future.
-			wp_send_json_error( 'schedule_changeset_needs_future_date' );
-		}
-
-		/*
-		 * Reset post date if we are publishing.
-		 */
-		if ( 'publish' === $changeset_status ) {
-			$changeset_date_gmt = '0000-00-00 00:00:00';
+		} elseif ( ! $is_updating_future_dated_changeset && 'future' === $changeset_status ) {
+			// Bail if date is not passed and existing date is not in future.
+			wp_send_json_error( 'not_future_date' );
 		}
 
 		$arg = array(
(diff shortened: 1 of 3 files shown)


src/wp-includes/class-wp-customize-manager.php
```diff
@@ -2034,28 +2034,53 @@ public function save() {
...
+		                                             ( mysql2date( 'U', $changeset_post->post_date_gmt, false ) > mysql2date( 'U', $now, false ) ) );
 
+		if ( isset( $changeset_date_gmt ) ) {
 			$is_future_dated = ( mysql2date( 'U', $changeset_date_gmt, false ) > mysql2date( 'U', $now, false ) );
 			if ( ! $is_future_dated ) {
-				wp_send_json_error( 'not_future_date', 400 ); // Only future dates are allowed.
+				if ( 'future' === $changeset_status ) {
+					wp_send_json_error( 'not_future_date', 400 ); // Only future dates are allowed.
+				} else {
+					$changeset_date_gmt = '0000-00-00 00:00:00';
```
westonruter: This will silently override a supplied past-date date to be the current time. Shouldn't this return an error as well here?

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -2034,28 +2034,53 @@ public function save() {
...
+			// Bail if date is not passed existing date is not in future.
+			wp_send_json_error( 'schedule_changeset_needs_future_date' );
 		}
 
-		$r = $this->save_changeset_post( array(
+		/*
+		 * Reset post date if we are publishing.
+		 */
+		if ( 'publish' === $changeset_status ) {
+			$changeset_date_gmt = '0000-00-00 00:00:00';
```
westonruter: It looks like `\WP_Customize_Manager::save_changeset_post()` needs to be updated to account for these reset-dates, specifically at:

```php
if ( $args['date_gmt'] ) {
	$post_array['post_date_gmt'] = $args['date_gmt'];
	$post_array['post_date'] = get_date_from_gmt( $args['date_gmt'] );
}
```

However, should not the supplying of the `0000-00-00 00:00:00` reset date be done _inside_ of `\WP_Customize_Manager::save_changeset_post()` instead of inside this Ajax handler?

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -2034,28 +2034,53 @@ public function save() {
 				}
 				$changeset_date_gmt = gmdate( 'Y-m-d H:i:s', $timestamp );
 			}
-			$now = gmdate( 'Y-m-d H:i:59' );
+		}
+
+		$changeset_post = get_post( $changeset_post_id );
```
westonruter: This should be removed because if `$changeset_post_id` is `null` then it will return the current global post.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -2034,28 +2034,53 @@ public function save() {
...
+		$now = gmdate( 'Y-m-d H:i:59' );
+		$is_valid_changeset_post_and_future_date = ( $changeset_post instanceof WP_Post
```
westonruter: Consider:

```php
$is_updating_future_dated_changeset = false;
if ( $changeset_post_id ) {
    $changeset_post = get_post( $changeset_post_id );
    $is_updating_future_dated_changeset = mysql2date( 'U', $changeset_post->post_date_gmt, false ) > mysql2date( 'U', $now, false ) );
}
```

src/wp-includes/class-wp-customize-manager.php
PatelUtkarsh: Yes, so we are only allowing future dates to be passed with any changeset status?

westonruter: A user should never be able to supply a date in the past. If they want to reset the date to *now* then the `date_gmt` should be supplied as `0000-00-00 00:00:00` or even `now`.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -2034,28 +2034,43 @@ public function save() {
...
+		}
+
+		if ( isset( $changeset_date_gmt ) ) {
 			$is_future_dated = ( mysql2date( 'U', $changeset_date_gmt, false ) > mysql2date( 'U', $now, false ) );
 			if ( ! $is_future_dated ) {
 				wp_send_json_error( 'not_future_date', 400 ); // Only future dates are allowed.
 			}
 
-			if ( ! $this->is_theme_active() && ( 'future' === $changeset_status || $is_future_dated ) ) {
+			if ( ! $this->is_theme_active() && $is_future_dated ) {
```
westonruter: Is this condition change fully correct? Should this even be expanded to: `( 'future' === $changeset_status || $is_future_dated || $is_updating_future_dated_changeset)`?

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -2034,20 +2034,6 @@ public function save() {
...
-				wp_send_json_error( 'not_future_date', 400 ); // Only future dates are allowed.
-			}
-
-			if ( ! $this->is_theme_active() && ( 'future' === $changeset_status || $is_future_dated ) ) {
-				wp_send_json_error( 'cannot_schedule_theme_switches', 400 ); // This should be allowed in the future, when theme is a regular setting.
-			}
-			$will_remain_auto_draft = ( ! $changeset_status && ( ! $changeset_post_id || 'auto-draft' === get_post_status( $changeset_post_id ) ) );
-			if ( $changeset_date && $will_remain_auto_draft ) {
-				wp_send_json_error( 'cannot_supply_date_for_auto_draft_changeset', 400 );
-			}
```
westonruter: @PatelUtkarsh I moved this into `\WP_Customize_Manager::save_changeset_post()` since logically this shouldn't be done in the Ajax handler but rather should be done for any time that `\WP_Customize_Manager::save_changeset_post()` is invoked by anyone.

PatelUtkarsh: 👍  Makes sense.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -2132,9 +2117,43 @@ function save_changeset_post( $args = array() ) {
 		$changeset_post_id = $this->changeset_post_id();
 		$existing_changeset_data = array();
 		if ( $changeset_post_id ) {
+			$existing_status = get_post_status( $changeset_post_id );
+			if ( 'publish' === $existing_status || 'trash' === $existing_status ) {
+				return new WP_Error( 'changeset_already_published' ); // @todo Add test for this.
```
westonruter: Oh, I suppose this _is_ already tested indirectly in the Ajax unit tests: https://github.com/xwp/wordpress-develop/pull/208/commits/89e553c3cca4064b1f2486349e208395991839c5#diff-a19659d3dd80459da2aaab2531b3de19

westonruter:
Now in trunk.

PR #209, opened 2016-11-29
#38948 - Make sure Sortables is initialized after the section has been expanded.

delawski:
Make sure Sortables is initialized after the section has been expanded to prevent `offset` issues.


westonruter:
Committed in r39376.

PR #210, opened 2016-11-30
Trac 38928: Default theme content - problem with slugs

westonruter:
https://core.trac.wordpress.org/ticket/38928


westonruter:
Now in trunk.

PR #211, opened 2016-12-05
Trac 39078: Theme starter content pages 404 if multiple auto-drafts were saved

westonruter:
https://core.trac.wordpress.org/ticket/39078


westonruter:
Now in trunk.

PR #213, opened 2017-01-26
Trac 31089: Customize: Changeset post_date does not clear when an auto-draft is updated

westonruter:
https://core.trac.wordpress.org/ticket/31089


westonruter:
Committed to trunk.

PR #218, opened 2017-03-07
Trac 39909: Make title behaviours consistent across all widgets on first load

westonruter:
https://core.trac.wordpress.org/ticket/39909


westonruter:
Committed to trunk in r40251.

PR #220, opened 2017-03-22
Trac #39087: Customizer: Add indicator when additional widget areas are hidden (2)

westonruter:
https://core.trac.wordpress.org/ticket/39087


westonruter:
Committed to trunk in r40330.

PR #221, opened 2017-04-10
WIP #37004 - Html5 / W3C Validation for the Posts page

mehigh:
https://core.trac.wordpress.org/ticket/37004


miina:
@mehigh 

Tested that the following list pages are not giving errors:
- Media
- Posts
- Plugins

Pages that gave an error:
- Comments list error: `Table columns in range 2…5 established by element td have no cells beginning in them.`;
- Users list: is giving one error: `Duplicate ID: changeit` -- (not sure if that's something to change);
- Pages list: is giving one error: `Duplicate ID: post_parent` (not sure if that's something to change);

Also found some HTML errors on other pages, such as General Settings page. Is this something that is out of scope for this ticket?

mehigh:
I've only looked at "Media" and "Posts". The reporter only complained about the Posts page having issues and I've globally applied the fixes for issues that were present on the Posts page, that's how Media got fixed as well.

mehigh:
Valuable feedback, Miina. Thank you.
I'll go further and address the other validation errors you've spotted.

mehigh:
@ Pages list: is giving one error: Duplicate ID: post_parent (not sure if that's something to change)

If there's no ID passed to the dropdown builder, it defaults to the name field - due to this:
if ( empty( $r['id'] ) ) {
		$r['id'] = $r['name'];
}

and because I can't access the $post->ID from 

{code}
$dropdown_args = array(
			'post_type'         => $post_type_object->name,
			'selected'          => $post->post_parent,
			'name'              => 'post_parent',
			'show_option_none'  => __( 'Main Page (no parent)' ),
			'option_none_value' => 0,
			'sort_column'       => 'menu_order, post_title',
		);
{code}
from

mehigh:
@miina This was refreshed. Thank you for your feedback.

src/wp-admin/includes/class-wp-media-list-table.php
```diff
@@ -678,7 +678,7 @@ private function _get_row_actions( $post, $att_title ) {
 				}
 			}
 			$actions['view'] = sprintf(
-				'<a href="%s" aria-label="%s" rel="permalink">%s</a>',
+				'<a href="%s" aria-label="%s" rel="bookmark">%s</a>',
```
westonruter: A caution here: there could be some CSS targeting `a[rel="permalink"]`.

mehigh: I'll check and update accordingly. I've updated all permalink occurrences by the bookmark variation, so it will be easy to update in the admin stylesheets. Thanks for the suggestion!

mehigh: The only results when searching for permalink in the css files were when it was used as a .permalink-structure class on a table.
I searched through the JS files as well, and we're covered confirming this doesn't break any existing functionality or style.

mehigh:
This was merged into core. Closing.

PR #222, opened 2017-04-10
#37013: "Toggle indicator" in pages have focus color while the same in widgets/menus have none

mehigh:
Major accessibility improvements in the widgets & menu pages.
https://core.trac.wordpress.org/ticket/37013


mehigh:
Scheduled for 4.8.1.

PR #225, opened 2017-04-17
Trac #40450: Add oembed/1.0/proxy endpoint

westonruter:
https://core.trac.wordpress.org/ticket/40450


timmyc:
I have some tests ready to push... should I fork this or can I get contrib status on this repo?

```
➜  public_html git:(trac-40450) git push origin trac-40450
ERROR: Permission to xwp/wordpress-develop.git denied to timmyc.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

westonruter:
@timmyc sorry, I neglected to give you write access. Should be good now.

timmyc:
Tests added in 64cccb7 - one of which actually fetches an embed from YouTube... I tried to look around and see if there was another test in the oembed suite that was doing this and nothing jumped out at me.  Not sure if having a test hit an external embed endpoint is the best thing to do.

westonruter:
@timmyc Instead of actually making the requests off to the external providers, I think it would probably be a better approach to add a `pre_http_request` filter in the unit test to preempt the request and mock the return value that the external service would provide. This will make the tests much faster and more reliable. So in your tests this could be done by just adding a `setUp` method that does the `add_filter( 'pre_http_request', … )` call and then all of the requests in the tests will automatically get mocked .

timmyc:
Thanks @westonruter - I had a hunch I was `_doing_it_wrong` there, thus the comment.  Mocking the request in c49bec7 now

tests/phpunit/tests/oembed/controller.php
```diff
@@ -18,6 +35,26 @@ public function setUp() {
 		$this->server = $wp_rest_server = new Spy_REST_Server();
 
 		do_action( 'rest_api_init', $this->server );
+		add_filter( 'pre_http_request', array( $this, 'mock_embed_request' ), 10, 3 );
+	}
+
+	public function tearDown() {
+		parent::tearDown();
+
+		remove_filter( 'pre_http_request', array( $this, 'mock_embed_request' ) );
```
westonruter: BTW, it's not necessary to do this. The filters already get reset in `tearDown`: https://github.com/WordPress/wordpress-develop/blob/8a2be374511b9002c93c25c16a113d24ae7bd61b/tests/phpunit/includes/testcase.php#L269-L292

tests/phpunit/tests/oembed/controller.php
```diff
@@ -18,6 +35,26 @@ public function setUp() {
...
+	public function mock_embed_request( $false, $arguments, $url ) {
+		// Mock request to YouTube Embed.
+		if ( $url === 'https://www.youtube.com/oembed?maxwidth=500&maxheight=750&url=https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3DOQSNhk5ICTI&format=json' ) {
+			return array(
+				'response' => array(
+					'code' => 200,
+				),
+				'body' => '{ "version": "1.0", "type": "video", "provider_name": "YouTube", "provider_url": "https://www.youtube.com", "thumbnail_width": 480, "width": 500, "thumbnail_height": 360, "html": "<iframe width=\"500\" height=\"375\" src=\"https://www.youtube.com/embed/OQSNhk5ICTI?feature=oembed\" frameborder=\"0\" allowfullscreen></iframe>", "author_name": "Yosemitebear62", "thumbnail_url": "https://i.ytimg.com/vi/OQSNhk5ICTI/hqdefault.jpg", "title": "Yosemitebear Mountain Double Rainbow 1-8-10", "height": 375 }',
+			);
+		}
```
westonruter: Also need to mock/short-circuit the request for `https://www.notreallyanoembedprovider.com/watch?v=OQSNhk5ICTI` to prevent an HTTP request from being made and slowing down the tests.

tests/phpunit/tests/oembed/controller.php
```diff
@@ -18,6 +35,26 @@ public function setUp() {
...
+	}
+
+	public function mock_embed_request( $false, $arguments, $url ) {
+		// Mock request to YouTube Embed.
+		if ( $url === 'https://www.youtube.com/oembed?maxwidth=500&maxheight=750&url=https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3DOQSNhk5ICTI&format=json' ) {
+			return array(
+				'response' => array(
+					'code' => 200,
+				),
+				'body' => '{ "version": "1.0", "type": "video", "provider_name": "YouTube", "provider_url": "https://www.youtube.com", "thumbnail_width": 480, "width": 500, "thumbnail_height": 360, "html": "<iframe width=\"500\" height=\"375\" src=\"https://www.youtube.com/embed/OQSNhk5ICTI?feature=oembed\" frameborder=\"0\" allowfullscreen></iframe>", "author_name": "Yosemitebear62", "thumbnail_url": "https://i.ytimg.com/vi/OQSNhk5ICTI/hqdefault.jpg", "title": "Yosemitebear Mountain Double Rainbow 1-8-10", "height": 375 }',
```
westonruter: I recommend doing `wp_json_encode()` on a regular PHP array.

tests/phpunit/tests/oembed/controller.php
```diff
@@ -18,6 +35,26 @@ public function setUp() {
...
+
+	public function tearDown() {
+		parent::tearDown();
+
+		remove_filter( 'pre_http_request', array( $this, 'mock_embed_request' ) );
+	}
+
+	public function mock_embed_request( $false, $arguments, $url ) {
+		// Mock request to YouTube Embed.
+		if ( $url === 'https://www.youtube.com/oembed?maxwidth=500&maxheight=750&url=https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3DOQSNhk5ICTI&format=json' ) {
```
westonruter: I suggest perhaps changing this to just look for the `OQSNhk5ICTI` substring, to make it more resilient. You might want to this a class constant even for re-use.

tests/phpunit/tests/oembed/controller.php
```diff
@@ -351,4 +397,54 @@ function test_get_oembed_endpoint_url_pretty_permalinks() {
...
+		$request = new WP_REST_Request( 'GET', '/oembed/1.0/proxy' );
+		$response = $this->server->dispatch( $request );
+
+		$this->assertEquals( 400, $response->get_status() );
+	}
+
+	public function test_proxy_with_invalid_oembed_provider() {
+		wp_set_current_user( self::$editor );
+		$request = new WP_REST_Request( 'GET', '/oembed/1.0/proxy' );
+		$request->set_param( 'url', 'https://www.notreallyanoembedprovider.com/watch?v=OQSNhk5ICTI' );
```
westonruter: Per above, this could be changed to not have `OQSNhk5ICTI` to prevent it from being mistaken for the YouTube request.

timmyc:
Thanks again for the feedback @westonruter, updated in ed786e7.

tests/phpunit/tests/oembed/controller.php
```diff
@@ -18,6 +36,35 @@ public function setUp() {
...
+						'thumbnail_height' => 360,
+						'html' => '<iframe width="500" height="375" src="https://www.youtube.com/embed/OQSNhk5ICTI?feature=oembed" frameborder="0" allowfullscreen></iframe>',
+						'author_name' => 'Yosemitebear62',
+						'thumbnail_url' => 'https://i.ytimg.com/vi/OQSNhk5ICTI/hqdefault.jpg',
+						'title' => 'Yosemitebear Mountain Double Rainbow 1-8-10',
+						'height' => 375,
+					)
+				),
+			);
+		}
```
westonruter: Shouldn't the else here be something like:

```
return array(
    'response' => array(
        'code' => 404,
    ),
    // ...
)
```

As otherwise, the oEmbed request for `https://www.notreallyanoembedprovider.com` will leak out and hit the network.

tests/phpunit/tests/oembed/controller.php
```diff
@@ -18,6 +36,35 @@ public function setUp() {
 		$this->server = $wp_rest_server = new Spy_REST_Server();
 
 		do_action( 'rest_api_init', $this->server );
+		add_filter( 'pre_http_request', array( $this, 'mock_embed_request' ), 10, 3 );
+	}
+
+	public function mock_embed_request( $false, $arguments, $url ) {
+		// Mock request to YouTube Embed.
+		if ( substr_count( $url, self::YOUTUBE_VIDEO_ID ) > 0 ) {
```
westonruter: I suggest using `strpos` instead of `substr_count`, as it's more commonly used in core (and it's also a tiny bit faster since it stops once finding the first occurance):

```php
if ( false !== strpos( $url, self::YOUTUBE_VIDEO_ID ) ) {
```

tests/phpunit/tests/oembed/controller.php
```diff
@@ -18,6 +36,35 @@ public function setUp() {
...
+						'version' => '1.0',
+						'type' => 'video',
+						'provider_name' => 'YouTube',
+						'provider_url' => 'https://www.youtube.com',
+						'thumbnail_width' => 480,
+						'width' => 500,
+						'thumbnail_height' => 360,
+						'html' => '<iframe width="500" height="375" src="https://www.youtube.com/embed/OQSNhk5ICTI?feature=oembed" frameborder="0" allowfullscreen></iframe>',
+						'author_name' => 'Yosemitebear62',
+						'thumbnail_url' => 'https://i.ytimg.com/vi/OQSNhk5ICTI/hqdefault.jpg',
```
westonruter: Interesting that the video ID is `3DOQSNhk5ICTI` and yet the URL contains `OQSNhk5ICTI` omitting the first `3D` part.

timmyc: wow that is interesting... since in the video widget we are simply using the video ID to build the thumbnail URL 🤔  maybe they limit it to a certain number of chars.

tests/phpunit/tests/oembed/controller.php
timmyc: Seems reasonable, shouldn't https://github.com/xwp/wordpress-develop/pull/225/files#diff-def8c9e660e5b02ad41bf045d6120612R153 prevent this from happening though?  Is there a way to monitor all network traffic generated by the test suite?

westonruter: Oh, good point. But actually `\WP_oEmbed::get_provider()` supports _discovery_. If you look at the function, you can see it takes a `discover` which will do a fallback request to see if the URL provided has oEmbed `link` element in its source. So actually I believe this would still result in an HTTP request.

As for how to monitor HTTP requests, a quick and dirty way is to just log out whenever the `pre_http_request` filter is applied. I tried adding `print "\nREQUESTED: $url\n";` to `\Test_oEmbed_Controller::mock_embed_request()` and you can see it tried to make a request for the unrecognized (not-builtin) provider: 

```
Installing...
Running as single site... To run multisite, use -c tests/phpunit/multisite.xml
Not running ajax tests. To execute these, use --group ajax.
Not running ms-files tests. To execute these, use --group ms-files.
Not running external-http tests. To execute these, use --group external-http.
PHPUnit 5.7.15 by Sebastian Bergmann and contributors.

..........S.......
REQUESTED: https://www.notreallyanoembedprovider.com/watch?v=awesome-cat-video
..                                              20 / 20 (100%)
REQUESTED: https://www.youtube.com/oembed?maxwidth=500&maxheight=750&url=https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3D3DOQSNhk5ICTI&format=json


Time: 1.44 seconds, Memory: 40.00MB
```

src/wp-includes/media.php
```diff
@@ -3423,6 +3424,7 @@ function wp_enqueue_media( $args = array() ) {
 			'audio' => ( $show_audio_playlist ) ? 1 : 0,
 			'video' => ( $show_video_playlist ) ? 1 : 0,
 		),
+		'oEmbedProxyUrl' => rest_url( 'oembed/1.0/proxy' ),
```
timmyc: awesome!  🎉

src/wp-includes/js/media-views.js
```diff
@@ -4646,7 +4646,13 @@ EmbedLink = wp.media.view.Settings.extend({
 			dataType: 'json',
 			context: this
 		})
-			.done( this.renderoEmbed )
+			.done( function( response ) {
+				this.renderoEmbed( {
+					data: {
+						body: response.html || ''
```
timmyc: if `response.html` is empty should we display something similar to `this.renderFail`?

westonruter: Perhaps. The reason why I did it this way is that `renderoEmbed` itself does similar checks on the `response` to obtain the `html`. If the request returns `200 OK` and if the `html` is empty, then maybe it is expected, or at least it would be in-line with what core is currently doing.

Aside: I did a double-tack when I saw `response.html` thinking it was some random filename in the source code 😄

timmyc: Gotcha, that makes sense for us to lean on the existing core checks.  It does seem odd that an embed would return an empty body though... but the internet is a strange place at times.

src/wp-includes/class-wp-oembed-controller.php
```diff
@@ -86,4 +111,53 @@ public function get_item( $request ) {
...
+	 *
+	 * @see WP_oEmbed::get_html()
+	 * @param WP_REST_Request $request Full data about the request.
+	 * @return WP_Error|array oEmbed response data or WP_Error on failure.
+	 */
+	public function get_proxy_item( $request ) {
+
+		$wp_oembed = _wp_oembed_get_object();
+		$url = $request['url'];
+		$args = $request->get_params();
```
westonruter: Thinking about this some more, I think we should actually add transient caching here. We can store the resulting `$data` in a transient for `oembed_ttl` seconds logic similar to what is found here: https://github.com/WordPress/wordpress-develop/blob/4.7.3/src/wp-includes/class-wp-embed.php#L195-L210

So the transient key should be a (truncated) hash of the MD5 hash of the serialized `$args` array.

By caching in a transient, then client-side applications that perhaps want to use oEmbeds on the frontend will not blow up their WordPress with a stampede of external requests spawned by visitors.

westonruter: > By caching in a transient, then client-side applications that perhaps want to use oEmbeds on the frontend will not blow up their WordPress with a stampede of external requests spawned by visitors.

Well, this isn't really relevant because it has an `edit_post` restriction and embeds would be embedded in the `content.rendered` anyway.

src/wp-includes/class-wp-oembed-controller.php
```diff
@@ -86,4 +111,53 @@ public function get_item( $request ) {
...
+	 * Checks if current user can make a proxy oEmbed request.
+	 *
+	 * @since 4.8.0
+	 * @access public
+	 *
+	 * @return true|WP_Error True if the request has read access, WP_Error object otherwise.
+	 */
+	function get_proxy_item_permissions_check() {
+
+		if ( ! current_user_can( 'edit_posts' ) ) {
```
westonruter: Is this the right capability check? It's basically what currently is used by the `parse-embed` admin-ajax request, but that's for the reason that the request mutates the postmeta of the associated post. In this case, there is no associated context post and no postmeta would be mutated.

timmyc: In the context of media preview, `edit_posts` seems to make sense since it covers the contributor role even.  Outside of someone trying to use this proxy for a pure JS theme which would potentially need un-authed support, it seems like `edit_posts` is a good capability to use here.

I suppose we could make the capability filterable, but still kind of feels overkill to me.

src/wp-includes/class-wp-oembed-controller.php
```diff
@@ -86,4 +111,66 @@ public function get_item( $request ) {
...
+
+		$data = $wp_oembed->fetch( $provider, $url, $args );
+		if ( false === $data ) {
+			return new WP_Error( 'oembed_no_provider_response', get_status_header_desc( 404 ), array( 'status' => 404 ) );
+		}
+
+		$data = (array) $data;
+
+		/** This filter is documented in wp-includes/class-wp-embed.php */
+		$ttl = apply_filters( 'oembed_ttl', DAY_IN_SECONDS, $url, $args, 0 );
```
westonruter: Note that the 3rd argument in the `oembed_ttl` filter is supposed to to be the post ID. Since this is outside a post context, there is no ID to supply and so that is why `0` is passed. This may make the filter unsuitable for use here, however.

tests/phpunit/tests/oembed/controller.php
nylen: Half-Life 3 confirmed.

src/wp-includes/class-wp-oembed-controller.php
nylen: I agree this could break existing code - I think `rest_oembed_ttl` would be a suitable alternative.

westonruter: Or even we could just skip having a filter altogether. We could add it later if there is a need.

tests/phpunit/tests/oembed/controller.php
nylen: More seriously, I think the video ID beginning with `3D` is simply invalid.  Note this is almost `%3D` - the URL-encoding of `=`.

- https://www.youtube.com/watch?v=3DOQSNhk5ICTI - YouTube lops off the `TI`, video not found
- https://www.youtube.com/watch?v=OQSNhk5ICTI - loads correctly

src/wp-includes/class-wp-oembed-controller.php
nylen: I think this is OK.  If we want to make something filterable here, I'd filter the result of the capability check instead, but this can always be done later.

src/wp-includes/class-wp-oembed-controller.php
```diff
@@ -86,4 +111,66 @@ public function get_item( $request ) {
...
+	/**
+	 * Checks if current user can make a proxy oEmbed request.
+	 *
+	 * @since 4.8.0
+	 * @access public
+	 *
+	 * @return true|WP_Error True if the request has read access, WP_Error object otherwise.
+	 */
+	function get_proxy_item_permissions_check() {
+
```
nylen: Unneeded newline here.

tests/phpunit/tests/rest-api/rest-schema-setup.php
```diff
@@ -43,6 +43,7 @@ public function test_expected_routes_in_schema() {
 			'/',
 			'/oembed/1.0',
 			'/oembed/1.0/embed',
+			'/oembed/1.0/proxy',
```
nylen: The new route should also be added below:

```php
array(
    'route' => '/oembed/1.0/proxy',
    'name'  => 'oembedProxy',
 ),
```

After rerunning the PHPUnit test suite, this will cause changes in a couple of the `wp-api.js` fixture files, which should also be included here.

tests/phpunit/tests/oembed/controller.php
```diff
@@ -18,6 +36,59 @@ public function setUp() {
...
+		$this->request_count += 1;
+
+		// Mock request to YouTube Embed.
+		if ( false !== strpos( $url, self::YOUTUBE_VIDEO_ID ) ) {
+			return array(
+				'response' => array(
+					'code' => 200,
+				),
+				'body' => wp_json_encode(
+					array(
```
nylen: It'd be nice to see this block aligned at the `=>`.

src/wp-includes/media.php
```diff
@@ -3414,6 +3414,7 @@ function wp_enqueue_media( $args = array() ) {
 		'captions'  => ! apply_filters( 'disable_captions', '' ),
 		'nonce'     => array(
 			'sendToEditor' => wp_create_nonce( 'media-send-to-editor' ),
+			'wpRestApi' => wp_create_nonce( 'wp_rest' ),
```
nylen: It'd be nice to see this block aligned at the `=>`.

src/wp-includes/js/media/views/embed/link.js
```diff
@@ -46,23 +45,25 @@ EmbedLink = wp.media.view.Settings.extend({
...
+			context: this
 		})
-			.done( this.renderoEmbed )
+			.done( function( response ) {
+				this.renderoEmbed( {
+					data: {
+						body: response.html || ''
+					}
+				} );
+			} )
```
nylen: This block looks to be identical to the one above it in `js/media-views.js`.  Can the version above just be re-used as a function here?

nylen:
One thing that will be made more difficult here: development of new embed endpoints, due to more aggressive caching.  In the past it's been possible to reset the caching by either switching to a new post or adding a filter on `oembed_ttl`.  It's not clear to me how this would be achieved using this endpoint.

tests/phpunit/tests/oembed/controller.php
```diff
@@ -9,6 +9,24 @@ class Test_oEmbed_Controller extends WP_UnitTestCase {
 	 * @var WP_REST_Server
 	 */
 	protected $server;
+	protected static $editor;
+	protected static $subscriber;
+	const YOUTUBE_VIDEO_ID = '3DOQSNhk5ICTI';
```
westonruter: Sure enough, as @nylen points out, this should be changed to just `OQSNhk5ICTI` per https://www.youtube.com/watch?v=OQSNhk5ICTI

And then `self::YOUTUBE_VIDEO_ID` can be used in place of `OQSNhk5ICTI` if so desired in the mock and assertions.

src/wp-includes/js/media/views/embed/link.js
westonruter: @nylen they look identical because they are identical 😄  The `link.js` here gets compiled into `media-views.js` via `grunt browserify`. So the duplication here is expected.

tests/phpunit/tests/oembed/controller.php
timmyc: Thanks for catching the double rainbow error @nylen

tests/phpunit/tests/oembed/controller.php
westonruter: See 3811180b360b77150b5c93089ccd01eb0de6f2c6

tests/phpunit/tests/rest-api/rest-schema-setup.php
westonruter: See 62448e0385ef04a2adaf5febfa95fe7a004755c1

westonruter:
> One thing that will be made more difficult here: development of new embed endpoints, due to more aggressive caching. In the past it's been possible to reset the caching by either switching to a new post or adding a filter on oembed_ttl. It's not clear to me how this would be achieved using this endpoint.

@timmyc more difficult to develop new embed endpoints or more difficult to develop _with_ this new oEmbed endpoint? Since this is a proxy endpoint, I didn't think that it would pose a difficulty since it's just relaying the data from another provider. But having a `rest_oembed_ttl` filter would alleviate your concern?

src/wp-includes/class-wp-oembed-controller.php
```diff
@@ -52,10 +52,35 @@ public function register_routes() {
 				),
 			),
 		) );
+
+		register_rest_route( 'oembed/1.0', '/proxy', array(
+			array(
+				'methods'  => WP_REST_Server::READABLE,
+				'callback' => array( $this, 'get_proxy_item' ),
+				'permission_callback' => array( $this, 'get_proxy_item_permissions_check' ),
+				'args'     => array(
```
aduth: Should `discover` be allowed as an `args` option (defaults `true`)?

https://github.com/WordPress/WordPress/blob/f1be128/wp-includes/class-oembed.php#L260-L261

src/wp-includes/class-wp-oembed-controller.php
```diff
@@ -86,4 +111,62 @@ public function get_item( $request ) {
...
+	 * Checks if current user can make a proxy oEmbed request.
+	 *
+	 * @since 4.8.0
+	 * @access public
+	 *
+	 * @return true|WP_Error True if the request has read access, WP_Error object otherwise.
+	 */
+	function get_proxy_item_permissions_check() {
+		if ( ! current_user_can( 'edit_posts' ) ) {
+			return new WP_Error( 'rest_forbidden_context', __( 'Sorry, you are not allowed to make proxied oEmbed requests.' ), array( 'status' => rest_authorization_required_code() ) );
```
aduth: In other controllers, the `_context` suffix is applied specific to the incoming request `context`. Since you're not testing that here, may be more appropriate to simplify to just `'rest_forbidden'`.

src/wp-includes/class-wp-oembed-controller.php
timmyc: Seems reasonable to support this here as well, especially if users want to disable the discovery request to speed response time.

src/wp-includes/class-wp-oembed-controller.php
westonruter: Good call. I think I copy-pasted this line when I saw “forbidden”.

src/wp-includes/class-wp-oembed-controller.php
westonruter: Sure, I don't see why the `discover` param shouldn't be supported.

nylen:
> @nylen more difficult to develop new embed endpoints or more difficult to develop with this new oEmbed endpoint?

Perhaps I was thinking ahead and/or misunderstanding a bit here, but I was considering the possibility that this new endpoint would be used for all front-end oEmbed requests.

westonruter:
@nylen got it. The purpose of this endpoint wouldn't be for front-end oEmbed requests, as those would be handled by `content.rendered` and the endpoint is restricted to those who can `edit_posts`. If it is used on the frontend, then it would be used in the context of frontend editing, and it would be the same scenario as integrating with the media modal as seen in this PR.

timmyc:
Added support for `discover` in 1755a2b, and fixed the error code.

src/wp-includes/class-wp-oembed-controller.php
timmyc: ✅  in 1755a2b

src/wp-includes/class-oembed.php
```diff
@@ -317,19 +317,18 @@ public static function _remove_provider_early( $format ) {
...
+	 * @since 4.8.0
 	 * @access public
 	 *
 	 * @param string       $url  The URL to the content that should be attempted to be embedded.
 	 * @param array|string $args Optional. Arguments, usually passed from a shortcode. Default empty.
-	 * @return false|string False on failure, otherwise the UNSANITIZED (and potentially unsafe) HTML that should be used to embed.
+	 * @return array|false oEmbed data array on success, false on failure.
 	 */
-	public function get_html( $url, $args = '' ) {
+	public function get_data( $url, $args = '' ) {
```
westonruter: Oh yes, this is great.

westonruter:
@swissspidy @nylen 4 tests are now failing: https://travis-ci.org/xwp/wordpress-develop/jobs/230930151#L607

swissspidy:
Yikes, thanks for the headsup. Will have a look ASAP.

westonruter:
Committed to `trunk`.

andyphillips82:
Hi Weston, 

Posted this in the forums, wondering if you have any thoughts.  I'm still very much in the investigation stage.
https://wordpress.org/support/topic/4-7-2-to-4-8-2-post-meta-spikes/

**Situation**: Transition from 4.7.2 to 4.8.2
**Problem**: post_meta response time spikes
**Suspected cause**: More selects hitting primary DB instead of replica
**Details**:  New Relic installed, Hyper DB installed.

What we're seeing is /feed/ hitting the primary DB more often.  
We believe because there are more update_post_meta coming from oEmbeds, but we're unsure. 
HyperDB will send the requests to primary if there are updates.  
Our primary DB responds slower as it's SSD vs our in memory replica DBs.

I've scoured the change logs from 4.7.2 to 4.8.2 but can't seem to find any related tickets other than this one.

I was wondering if you had any ideas?  I can't see why your changes would have any related impact, but I'm not 100% solid on the oEmbed lifecycle.

Thanks
Andy

andyphillips82:
Ultimately it touches the get_html and get_data functions, which is common to both.

westonruter:
I see. @swissspidy Does anything come to mind where the changes here would cause such a problem?

swissspidy:
The one thing that just caught my eye is that the `$args = wp_parse_args( $args );` line was moved from `get_html()` to `get_data()`. This means anyone hooking into the `pre_oembed_result` filter won't have the parsed args.

Also, `if ( false === $data ) { return false; } return $data;` is a bit silly.

@andyphillips82 Do you see any differences when you put that `wp_parse_args()` line back to the beginning of `get_html()`?

If we introduced a regression here, the best thing to do is to create a new ticket on Trac.

andyphillips82:
I'm out of the office today.  Working in my off time cause this one is driving me crazy :)

I'll try that tomorrow and report back!  It's intermittent so might take a few days to confirm either way.

andyphillips82:
Gah, I reverted the entire class-oembed.php to 4.7.2 and we're still seeing the spikes.

False alarm, appreciate the time guys.

andyphillips82:
Just an FYI, I commented out update_post_meta :260, :261, :263 in class-wp-embed.php and the issue goes away.

Since 4.8.2, we're seeing an increase in post_meta_updates and increased external traffic to Twitter.   Again, not sure if it's related to this at all, but it's very peculiar.

PR #227, opened 2017-05-05
Add media widgets

obenland:
Unit tests still need attention.


obenland:
Sounds good

celloexpressions:
I haven't had a chance to test this since the widgets were split out from a single media widget. Testing now, I found the following issues:
- Incorrect spacing between the placeholder and the buttons (2em total, instead of 1em = 13px). Remove top margin from buttons and add bottom margin on preview, or remove bottom margin on placeholder to fix.
- Widget descriptions should not use verbs. Currently, all of the new widgets start with "displays," while all existing widgets but one skip the verb (ex. "A calendar that displays your latest posts"). They should be changed to, for example, "An image from your media library or a URL." Note that this same issue was previously fixed for the single media widget.
- No warning or error when using the wrong type of URL for a given widget. I accidentally pasted a YouTube URL into the audio widget and there were no warnings or errors, but it then shows the link and "download file". There needs to be feedback about why it's not working for this situation or, even better, the widget should adapt to the appropriate widget type based on the content used. For people who frequently use YouTube for music, the audio widget seems to suggest that a YouTube URL would work but then it gives a very odd experience. Perhaps for external links that can oembed, we can embed the content regardless of what it is, even if it's the "wrong" widget. That thought also questions whether this functionality belongs as part of media-specific widgets or is more appropriate as a standalone widget. My initial thought is that we should keep it for media-specific, but eliminate validation constraints on source, and also add a generic "embed" widget that can embed anything that can oembed. That would facilitate featuring internal or external posts as well as a broader array of content types since posts support oembed.
- Missing clickable placeholders and clickable images to edit, see related trac ticket. I would think that it would be a lot easier to do it right when we introduce this feature rather than continuing a broken pattern.
- The distinction between "edit image" and "change image" is confusing, especially because both modals provide the same options besides the advanced toggle. This applies to both hosted and external images. Recommendation: eliminate the separate "edit image" button and modal, and rethink the "change image" label to encompass the ability to edit properties like the caption.
- Ridiculously-narrow inputs (select for dates and search input) in `.media-toolbar` in the media modal. Not sure if it's specific to the widgets' use of the modal but shouldn't be particularly hard to fix.
- The audio widget seems to open a modal titled "select video" for me. Which is part of why I tried trowing a youtube URL into it.
- On some themes, audio widgets fail to initialize with MediaElement.js and instead show a generic browser player.
- Default themes that customize MediaElement.js players need to be updated so that players display the same both in posts and in widgets. The end result is quite jarring otherwise, as icons, colors, etc. are different. Theme authors should be notified as soon as possible that they need to take these new widgets into consideration, so that they have time to test and update prior to release.
- Multiple bundled themes either strip out or completely restyle `ol` and `ul` elements in widget areas, producing a very odd experience when the visual text widget offers bulletted and ordered list buttons that just don't work, and typing bullets or numbers converts them to list elements automatically and prevents them from showing up properly on the frontend. This area needs a lot more testing on different themes as well as strategies to produce reasonable results within widgets. Many themes do not intend for widgets to contain "content" like this, and anything that has an actual button in the mini-editor should work for all users without jumping through hoops.

In summary, I do not feel that this is quite ready for merge. Since it introduces a significant amount of UI, at least a few user tests should be conducted prior to finalizing and merging this implementation. Additionally, it seems like more thought should go into the ways that this content is structured and implications for forward-compatibility. I have not thoroughly tested, but these are my findings after about 20 minutes of testing.

melchoyce:
@celloexpressions: made some issues based on your feedback:
- https://github.com/xwp/wp-core-media-widgets/issues/148
- https://github.com/xwp/wp-core-media-widgets/issues/149
- https://github.com/xwp/wp-core-media-widgets/issues/150
- https://github.com/xwp/wp-core-media-widgets/issues/151
- https://github.com/xwp/wp-core-media-widgets/issues/152
- https://github.com/xwp/wp-core-media-widgets/issues/153

> The audio widget seems to open a modal titled "select video" for me. Which is part of why I tried trowing a youtube URL into it.

I'm unable to replicate this — are you still experiencing it? 

> On some themes, audio widgets fail to initialize with MediaElement.js and instead show a generic browser player.

Can you give us an example of a theme where this is happening?

> Default themes that customize MediaElement.js players need to be updated so that players display the same both in posts and in widgets. The end result is quite jarring otherwise, as icons, colors, etc. are different.

Do you know which default themes customizer the media player? Some screenshots would help here, if you can.

> Multiple bundled themes either strip out or completely restyle ol and ul elements in widget areas, producing a very odd experience when the visual text widget offers bulletted and ordered list buttons that just don't work, and typing bullets or numbers converts them to list elements automatically and prevents them from showing up properly on the frontend. This area needs a lot more testing on different themes as well as strategies to produce reasonable results within widgets. Many themes do not intend for widgets to contain "content" like this, and anything that has an actual button in the mini-editor should work for all users without jumping through hoops.

This will likely be easier for theme authors to test once we're in trunk and running a beta. We still have a couple weeks so theme authors should have some time to make updates based on the new widgets.

Thanks for testing!

src/wp-includes/widgets/class-wp-widget-media-video.php
```diff
@@ -0,0 +1,250 @@
...
+			$schema[ $video_extension ] = array(
+				'type' => 'string',
+				'default' => '',
+				'format' => 'uri',
+				/* translators: placeholder is video extension */
+				'description' => sprintf( __( 'URL to the %s video source file' ), $video_extension ),
+			);
+		}
+
+		// TODO: height and width?
```
westonruter: This can be removed now.

westonruter:
Committed to trunk in https://core.trac.wordpress.org/changeset/40640

PR #230, opened 2017-05-11
Add visual text widget

westonruter:
Splits out text widget from media widget PR (#227).

See https://core.trac.wordpress.org/ticket/35243


tests/phpunit/tests/widgets/test-class-wp-widget-text.php
```diff
@@ -0,0 +1,133 @@
...
+		$instance['filter'] = true;
+		ob_start();
+		$widget->widget( $args, $instance );
+		$output = ob_get_clean();
+		$this->assertContains( '<p>', $output );
+		$this->assertContains( '<br />', $output );
+
+		$instance['filter'] = 'content';
+		ob_start();
+		$widget->widget( $args, $instance );
```
westonruter: @obenland I think we should also check here to make sure that the `widget_text_content` filter is applying, adding a filter handler an ensuring it is passed the expected args and that it is called.

tests/phpunit/tests/widgets/test-class-wp-widget-text.php
```diff
@@ -0,0 +1,133 @@
...
+
+		ob_start();
+		$widget->widget( $args, $instance );
+		$output = ob_get_clean();
+		$this->assertNotContains( '<p>', $output );
+		$this->assertNotContains( '<br />', $output );
+
+		$instance['filter'] = true;
+		ob_start();
+		$widget->widget( $args, $instance );
```
westonruter: @obenland Likewise, I think we should check here to make sure that the `widget_text` filter is applies in every case, adding a filter handler an ensuring it is passed the expected args and that it is called. And also to verify in the case of `filter` not being `content` that the filter does _not_ apply.

tests/phpunit/tests/widgets/test-class-wp-widget-text.php
```diff
@@ -0,0 +1,133 @@
...
+		$this->assertSame( $result, $expected );
+	}
+
+	/**
+	 * Test enqueue_admin_scripts method.
+	 *
+	 * @covers WP_Widget_Text::enqueue_admin_scripts
+	 */
+	function test_enqueue_admin_scripts() {
+		set_current_screen( 'widgets.php' );
```
westonruter: This is going to need the same `clean_up_global_scope` method as implemented by other tests to reset the `$wp_scripts` and `$wp_styles` before the test.

tests/phpunit/tests/widgets/test-class-wp-widget-text.php
obenland: > adding a filter handler an ensuring it is passed the expected args

Is there an example you can point me to for this? Not sure how to attack that…

westonruter: Yes, see https://github.com/xwp/wp-core-media-widgets/blob/master/tests/phpunit/test-class-wp-widget-media.php#L321

tests/phpunit/tests/widgets/test-class-wp-widget-text.php
```diff
@@ -100,6 +166,9 @@ function test_update() {
 		wp_get_current_user()->add_cap( 'unfiltered_html' );
 		$expected['text'] = '<script>alert( "Howdy!" );</script>';
 		$result = $widget->update( $expected, $instance );
+		if ( defined( 'WP_TESTS_MULTISITE') && WP_TESTS_MULTISITE ) {
```
obenland: @westonruter This feels so hacky. Is there a better way to check if we're in multisite?

westonruter: What about checking `current_user_can( 'unfiltered_html' )` instead?

obenland: Found a work-around, c3268fb. Test still fails on remote :(

westonruter:
Committed to `trunk` in [r40631](https://core.trac.wordpress.org/changeset/40631).

PR #231, opened 2017-05-25
#39221 Customize: Prevent setting previewing logic when settings will be saved

miina:
https://core.trac.wordpress.org/ticket/39221 

Implement `WP_Customize_Manager::settings_previewed()`.

Prevents setting preview in case of `customize_save` and adds an additional param `WP_Customize_Manager::settings_previewed` for other possible cases, e.g. in case of https://github.com/WP-API/wp-api-customize-endpoints when publishing a changeset via REST endpoint.


src/wp-includes/class-wp-customize-manager.php
```diff
@@ -193,6 +193,15 @@
 	protected $messenger_channel;
 
 	/**
+	 * If to skip previewing settings.
+	 *
+	 * @since 4.?.?
+	 * @access protected
+	 * @var bool
+	 */
+	protected $skip_setting_preview;
```
westonruter: I wonder if we should let this just be `$should_preview_settings` so it is named the same as the method, and then reverse the logic.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -264,9 +274,14 @@ public function __construct( $args = array() ) {
...
 		}
 
+		if ( ! isset( $args['skip_setting_preview'] ) ) {
+			$args['skip_setting_preview'] = false;
+		}
+
 		$this->original_stylesheet = get_stylesheet();
 		$this->theme = wp_get_theme( $args['theme'] );
 		$this->messenger_channel = $args['messenger_channel'];
+		$this->skip_setting_preview = true === $args['skip_setting_preview'];
```
miina: Shouldn't we validate if the value is actually `true` and not just something? (If I understood the suggestion correctly.)

src/wp-includes/class-wp-customize-manager.php
miina: For some reason `$skip_setting_preview` feels suitable here since this variable theoretically should be used only if the goal is to skip the preview. However, no strong preferences, so if you think that `$should_preview_settings` makes more sense or is more clear, happy to change that.

miina: Although yes, it would be consistent with `WP_Customize_Manager::should_preview_settings()` this way and thus more clear as well.

src/wp-includes/class-wp-customize-manager.php
westonruter: I think it's fine to just check if the value is truthy, and `! empty()` in that case will check if it is set. But not a big deal.

src/wp-includes/class-wp-customize-manager.php
westonruter: Also we should consider the existing getters and setters. Specifically, there is a `changeset_uuid` arg when constructing `WP_Customize_Manager` and then there is a `WP_Customize_Manager::changeset_uuid()` to read this back out. So I think the method name should be consistent with the arg that is passed in.

Let's also consider changing the arg and the method name to be just`settings_previewed`.

src/wp-includes/class-wp-customize-manager.php
miina: Okay, sure. Was thinking of cases of `"false"` or `"0"` when `! empty()` would be `true`.

westonruter:
68670c6 incorporates the same testing logic from https://github.com/xwp/wp-customize-snapshots/pull/137

westonruter:
Committed to `trunk` r41205.

PR #234, opened 2017-06-12
WP Core #40986: Add accessibility mode support to both Text widget and media widgets

westonruter:
https://core.trac.wordpress.org/ticket/40986


westonruter:
Unit tests need to be updated;

westonruter:
Committed in to `trunk` in r40941.

PR #235, opened 2017-06-23
#40951: Add legacy mode to Text widget

westonruter:
https://core.trac.wordpress.org/ticket/40951


src/wp-includes/widgets/class-wp-widget-text.php
```diff
@@ -53,6 +53,124 @@ public function _register() {
...
+
+		$doc = new DOMDocument();
+		$doc->loadHTML( sprintf(
+			'<html><head><meta charset="%s"></head><body>%s</body></html>',
+			esc_attr( get_bloginfo( 'charset' ) ),
+			$instance['text']
+		) );
+		$body = $doc->getElementsByTagName( 'body' )->item( 0 );
+
+		// See $allowedposttags.
```
coreymckrill: Why not just use `wp_kses_allowed_html( 'post' )` here?

westonruter: Good question. It's because we are looking for a subset of what is allowed, specifically what TineMCE won't munge.

westonruter: For example, TineMCE will munge `p` tags.

westonruter: And if there is a `span` or a `div` there is a good chance the HTML is “advanced” and TinyMCE shouldn't be loaded.

afercia:
It would be great to review all the links with `href="#"`, since they probably should be actual `button` elements. The problem with them is that, either they look like buttons or not, they're still reported as links by software reading the page. For example, they will be listed in the links list. If they perform an action, they should be buttons. They can be styled with the proper CSS classes `.button`, or `.button-link` (if they have to look like links).

I see a few of them, for example:
`<a class="add-widget" href="#">Add a Widget</a>`
`<a class="close" href="#"><?php _e( 'Dismiss' ); ?></a>`


I guess this is probably legacy, but I'm not sure why it should be a h3. Pointers are sort of "floating" panels and they may appear in a position where h3 doesn't make sense. This should be addressed for all pointers though, so I guess it's a bit out of the scope here.
`<h3><?php _e( 'New Custom HTML Widget' ); ?></h3>`

afercia:
@westonruter I've tested the pointer a11y improvements in Safari10+VoiceOver and Firefox+NVDA. The message is correctly announced and everything works fine. Thanks!

Worth noting this could be simplified using `role="alert"`, or  `role="alertdialog"`, or  `aria-live="assertive"` directly on the pointer wrapper element, but support is inconsistent across browsers and screen readers, while `wp.a11y.speak()` just works.

westonruter:
@afercia thanks for the review! 

> It would be great to review all the links with `href="#"`, since they probably should be actual `button` elements.
> I guess this is probably legacy, but I'm not sure why it should be a h3. Pointers are sort of "floating" panels and they may appear in a position where h3 doesn't make sense. This should be addressed for all pointers though, so I guess it's a bit out of the scope here.

Yes, the markup used here is just re-using what is in core:

* https://github.com/WordPress/wordpress-develop/blob/4.8.0/src/wp-includes/js/wp-pointer.js#L18
* https://github.com/WordPress/wordpress-develop/blob/4.8.0/src/wp-includes/css/wp-pointer.css#L11-L21

So we should indeed revisit this in the scope of another ticket, since it would impact all pointers.

westonruter:
@afercia question: when you have `role=alert` and other such attributes, does this cause screen readers to read them once the element transitions from `display:none` to `display:block` (regardless of whether the `hidden` attribute sticks around)? Or do screen readers read them only once you dynamically add the `role=alert` attribute to the element?

westonruter:
Committed to `trunk`.

afercia:
> Yes, the markup used here is just re-using what is in core:

OK thought to improve it since this is new code introduced in core, we can postpone.

> does this cause screen readers to read them once the element transitions from display:none to display:block

theoretically yes, because [alert](https://www.w3.org/TR/wai-aria/roles#alert) and [alertdialog](https://www.w3.org/TR/wai-aria/roles#alertdialog) are processed as assertive live regions but there are subtleties and [different interpretations](https://bugzilla.mozilla.org/show_bug.cgi?id=788428) by screen reader vendors so the implementation is not consistent. I wouldn't recommend to enter in that chaos :)

westonruter:
Remaining changes committed to trunk in https://core.trac.wordpress.org/changeset/41070

PR #237, opened 2017-07-05
#40974: Updated text widget do not save text (when pasting)

westonruter:
https://core.trac.wordpress.org/ticket/40974


src/wp-admin/js/widgets/text-widgets.js
```diff
@@ -184,15 +184,28 @@ wp.textWidgets = ( function( $ ) {
...
 				editor.on( 'NodeChange', _.debounce( triggerChangeIfDirty, changeDebounceDelay ) );
-				editor.on( 'blur', function() {
+
+				/*
+				 * Note that originally the 'blur' event was used below. It had to be
+				 * replaced with 'SaveContent' because the 'blur' event does not trigger
+				 * when the user switches to the Text tab.
+				 */
+				editor.on( 'SaveContent', function() {
 					control.editorFocused = false;
```
westonruter: I can see this `editorFocused` change is wrong here. It needs to be moved up to `blur` handler.

PR #238, opened 2017-07-06
#41252: Check return value of WP_Customize_Manager::get_changeset_post_data()

westonruter:
https://core.trac.wordpress.org/ticket/41252


westonruter:
Committed to `trunk` in r41012.

PR #239, opened 2017-07-12
#41299: oEmbed proxy fails to forward maxwidth and maxheight params

westonruter:
https://core.trac.wordpress.org/ticket/41299


westonruter:
Committed to `trunk` in https://core.trac.wordpress.org/changeset/41047

PR #240, opened 2017-07-19
Trac #40974: Updated text widget do not save text (when using paste)

westonruter:
Supersedes #237.

https://core.trac.wordpress.org/ticket/40974


westonruter:
Committed to `trunk` in https://core.trac.wordpress.org/changeset/41094

PR #242, opened 2017-07-26
Support YouTube Embed URLs

timmyc:
This PR adds support for YouTube embed URL's in both the media modal within the editor, and the Video Widget.

To test, use a YouTube embed URL in both places `https://www.youtube.com/embed/ea2WoUtbzuw` and verify the oembed proxy preview is shown properly, and the URL works in both the editor and the video widget.

### DIFF
--- a/src/wp-admin/js/widgets/media-widgets.js
+++ b/src/wp-admin/js/widgets/media-widgets.js
@@ -145,7 +145,7 @@ wp.mediaWidgets = ( function( $ ) {
 					 * @returns {void}
 					 */
 					fetch: function() {
-						var embedLinkView = this, fetchSuccess, matches, fileExt, urlParser; // eslint-disable-line consistent-this
+						var embedLinkView = this, fetchSuccess, matches, fileExt, urlParser, url, re, youTubeEmbedMatch; // eslint-disable-line consistent-this
 
 						if ( embedLinkView.dfd && 'pending' === embedLinkView.dfd.state() ) {
 							embedLinkView.dfd.abort();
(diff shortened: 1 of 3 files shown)


src/wp-includes/js/media-views.js
```diff
@@ -4624,16 +4623,24 @@ EmbedLink = wp.media.view.Settings.extend({
...
+		if ( $('#embed-url-field').val() !== url ) {
 			return;
 		}
 
 		if ( this.dfd && 'pending' === this.dfd.state() ) {
 			this.dfd.abort();
 		}
 
+		// Support YouTube embed urls, since they work once in the editor.
+		var re = /https?:\/\/www\.youtube\.com\/embed\/([^/]+)/;
```
westonruter: This is going to fail JSHint because the `var` is not at the top of the function.

src/wp-admin/js/widgets/media-widgets.js
```diff
@@ -190,10 +190,20 @@ wp.mediaWidgets = ( function( $ ) {
 							return;
 						}
 
+						// Support YouTube embed links.
+						url = embedLinkView.model.get( 'url' );
+						re = /https?:\/\/www\.youtube\.com\/embed\/([^/]+)/;
```
westonruter: Are there ever any query args on these URLs?

timmyc: Not in my testing, and they are always displayed with `www`.  FWIW the regex is mostly from: https://github.com/xwp/wordpress-develop/blob/master/src/wp-includes/embed.php#L190

src/wp-includes/js/media-views.js
westonruter: Humm JSHint isn't complaining. Anyway, WP code style says the `var` should be grouped at the top of the function: https://make.wordpress.org/core/handbook/best-practices/coding-standards/javascript/#declaring-variables-with-var

westonruter:
Committed to `trunk` in r41179.

PR #243, opened 2017-07-28
#41274: Expose link input in widget

timmyc:
This branch explores a design that displays the `link_url` input form for the image widget in the widget itself to make it more discoverable.

See mockups: https://cloudup.com/cTX3XurZQBJ

### DIFF
--- a/src/wp-admin/css/widgets.css
+++ b/src/wp-admin/css/widgets.css
@@ -152,10 +152,6 @@
 	border: 1px solid #f00;
 }
 
-.media-widget-extra {
-	text-align: left;
-}
-
 .media-widget-image-link {
 	margin: 1em 0;
 }
--- a/src/wp-admin/css/widgets.css
+++ b/src/wp-admin/css/widgets.css
@@ -152,6 +152,14 @@
 	border: 1px solid #f00;
 }
 
+.media-widget-extra {
+	text-align: left;
+}
+
+.media-widget-image-link {
+	margin: 1em 0;
+}
+
 /* Widget Dragging Helpers */
 .widget.ui-draggable-dragging {
 	min-width: 100%;
--- a/src/wp-admin/js/widgets/media-image-widget.js
+++ b/src/wp-admin/js/widgets/media-image-widget.js
@@ -37,13 +37,13 @@
 
 			previewContainer = control.$el.find( '.media-widget-preview' );
 			previewTemplate = wp.template( 'wp-media-widget-image-preview' );
-			previewContainer.html( previewTemplate( _.extend( control.previewTemplateProps.toJSON() ) ) );
+			previewContainer.html( previewTemplate( control.previewTemplateProps.toJSON() ) );
 
 			linkInput = control.$el.find( '.link' );
 			if ( ! linkInput.is( document.activeElement ) ) {
 				fieldsContainer = control.$el.find( '.media-widget-fields' );
 				fieldsTemplate = wp.template( 'wp-media-widget-image-fields' );
-				fieldsContainer.html( fieldsTemplate( _.extend( control.previewTemplateProps.toJSON() ) ) );
+				fieldsContainer.html( fieldsTemplate( control.previewTemplateProps.toJSON() ) );
 			}
 		},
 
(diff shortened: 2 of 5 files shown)


src/wp-includes/widgets/class-wp-widget-media.php
```diff
@@ -404,6 +404,8 @@ public function render_control_template_scripts() {
 					<?php echo esc_html( $this->l10n['add_media'] ); ?>
 				</button>
 			</p>
+			<div class="media-widget-extra">
```
westonruter: Maybe instead of “extra” we should refer to this as “additional-fields”? Or maybe even just “fields”?

src/wp-admin/js/widgets/media-widgets.js
```diff
@@ -513,6 +513,18 @@ wp.mediaWidgets = ( function( $ ) {
 				});
 			});
 
+			// Update extra attributes.
+			control.$el.on( 'input change', '.extra', function updateExtra() {
```
westonruter: Instead of targeting every `.extra` field, I think this should specifically be `.link` or `[data-attr="link_url"]`. Then there would be no need for the `data-attr` lookup below or the `( attr === 'link_url' )` check since it would only ever be for the link URL field.

src/wp-admin/js/widgets/media-image-widget.js
```diff
@@ -134,6 +141,7 @@
 			url = control.model.get( 'url' );
 			mediaFrameProps = component.MediaWidgetControl.prototype.mapModelToPreviewTemplateProps.call( control );
```
westonruter: Might as well rename `mediaFrameProps` to `previewTemplateProps` since that is what is returned.

src/wp-includes/widgets/class-wp-widget-media-image.php
```diff
@@ -307,10 +307,17 @@ public function render_control_template_scripts() {
 		parent::render_control_template_scripts();
 
 		?>
+		<script type="text/html" id="tmpl-wp-media-widget-image-extra">
```
westonruter: Per below, I think `fields` makes more sense than `extra`.

src/wp-includes/widgets/class-wp-widget-media.php
timmyc: makes sense.

src/wp-admin/js/widgets/media-widgets.js
timmyc: yeah my original thinking here was trying to set the stage for additional "fields" that might be used in this manner in other widgets.  I'm fine targeting specific for this use case and adding in a more general approach if/when other widgets add fields like this.

timmyc:
@westonruter made some updates based upon your feedback.

westonruter:
@timmyc I noticed a few more issues related to changing the the link type and link URL across the edit frame, select frame, and extra field. I could sometimes get this to happen:

image

And sometimes when changing the link type in the edit frame, it would cause the new link URL field to clear out. Also when switching the selected media, sometimes the URL would not update as I switched. And other times, when I'd change the custom URL it would not show up in the media select frame.

Hard to pinpoint and reproduce the steps to reproduce the issues I found, but after tweaking things for awhile I got it to a state where I wasn't able to encounter any further oddities: fbdb3e0d117454edc3665f246ad254e91b37385c

If that all looks and tests well for you, please confirm and then I'll commit.

/cc @melchoyce

src/wp-includes/widgets/class-wp-widget-media-image.php
```diff
@@ -307,10 +307,17 @@ public function render_control_template_scripts() {
 		parent::render_control_template_scripts();
 
 		?>
+		<script type="text/html" id="tmpl-wp-media-widget-image-fields">
+			<# var elementIdPrefix = 'el' + String( Math.random() ) + '_'; #>
+			<# if ( data.url ) { #>
+			<p class="media-widget-image-link">
+				<label for="{{ elementIdPrefix }}linkUrl"><?php esc_html_e( 'Link to:' ); ?></label>
+				<input id="{{ elementIdPrefix }}linkUrl" type="text" class="widefat link extra" value="{{ data.link_url }}" placeholder="http://">
```
westonruter: Oh, shouldn't this be `type="url"`?

src/wp-admin/css/widgets.css
```diff
@@ -152,6 +152,14 @@
 	border: 1px solid #f00;
 }
 
+.media-widget-extra {
```
westonruter: Is this a dead selector now?

src/wp-includes/widgets/class-wp-widget-media-image.php
westonruter: Is the `extra` class being used now?

src/wp-admin/js/widgets/media-widgets.js
```diff
@@ -513,6 +513,26 @@ wp.mediaWidgets = ( function( $ ) {
 				});
 			});
 
+			// Update link_url attribute.
+			control.$el.on( 'input change', '.link', function updateExtra() {
```
westonruter: This should probably be updateLinkUrl

timmyc:
Thanks for the updates there @westonruter - makes perfect sense that the linkUrl should trigger a preview update - good catch there.

westonruter:
Committed to `trunk` in r41252.

PR #244, opened 2017-08-03
#41540: Ensure that extended Text widgets are initialized with JS

westonruter:
https://core.trac.wordpress.org/ticket/41540


westonruter:
Committed to trunk in r41245.

PR #245, opened 2017-08-04
Trac 39692: Fix missing assignment of menus on theme switch

westonruter:
https://core.trac.wordpress.org/ticket/39692

### DIFF
--- a/src/wp-includes/nav-menu.php
+++ b/src/wp-includes/nav-menu.php
@@ -1036,7 +1036,7 @@ function _wp_delete_customize_changeset_dependent_auto_drafts( $post_id ) {
 function _wp_menus_changed() {
 	$old_nav_menu_locations = get_option( 'theme_switch_menu_locations', array() );
 	$new_nav_menu_locations = get_nav_menu_locations();
-	$remapped_nav_menu_locations = wp_get_remapped_nav_menu_locations( $new_nav_menu_locations, $old_nav_menu_locations );
+	$remapped_nav_menu_locations = wp_map_nav_menu_locations( $new_nav_menu_locations, $old_nav_menu_locations );
 	set_theme_mod( 'nav_menu_locations', $remapped_nav_menu_locations );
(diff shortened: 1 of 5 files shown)


src/wp-includes/nav-menu.php
```diff
@@ -1026,3 +1026,87 @@ function _wp_delete_customize_changeset_dependent_auto_drafts( $post_id ) {
...
+			if ( ! empty( $old_nav_menu_locations ) ) {
+				/*
+				 * If old and new theme both have locations that contain phrases
+				 * from within the same group, make an educated guess and map it.
+				 */
+				$common_slug_groups = array(
+					array( 'header', 'main', 'navigation', 'primary', 'top' ),
+					array( 'bottom', 'footer', 'secondary', 'subsidiary' ),
+					array( 'social' ),
+					// TODO: Find a second slug or remove, since locations with same slug are already mapped.
```
westonruter: @obenland what needs to be done here?

src/wp-includes/nav-menu.php
```diff
@@ -1026,3 +1026,87 @@ function _wp_delete_customize_changeset_dependent_auto_drafts( $post_id ) {
...
+										if ( false !== stripos( $location, $slug ) || false !== stripos( $slug, $location ) ) {
+											
+											// Make sure this location wasn't mapped and removed previously.
+											if ( ! empty( $old_nav_menu_locations[ $location ] ) ) {
+												
+												// We have a match that can be mapped!
+												$new_nav_menu_locations[ $new_location ] = $old_nav_menu_locations[ $location ];
+												
+												// Remove the mapped location so it can't be mapped again.
+												unset( $old_nav_menu_locations[ $location ] );
```
westonruter: @obenland I recall there being some unexpected behaviors when modifying an array while iterating over it. Is this why you have the `Make sure this location wasn't mapped and removed previously` condition above?

westonruter:
Committed to `trunk` in r41237.

PR #248, opened 2017-08-17
Fixes for MEJS in WP 4.9

rafa8626:
https://core.trac.wordpress.org/ticket/41640 

Also fixes:
- Accessibility for volume control
- Progress bar tooltip positioning and dimensions correctly set for media with long duration
- Better codec checking
- Minor CSS fix to avoid flickering effect on volume control


rafa8626:
@westonruter This is the last round I have to submit. At this point, I'll just be pending on any tickets coming from core.trac.wordpress.org and submit PRs for them

westonruter:
I'm curious why there is no version change with this.

westonruter:
It appears that the changes here would be part of a `4.2.6` release? Are you wanting to get the changes committed into WordPress core prior to releasing the new version to ensure it works in WordPress core? I should think it should go the other way around.

rafa8626:
The thing is I'm not maintaining MEJS anymore. I'm just wrapping things for WP and that's it. I could release the next version, but I'm not sure when I would do it

rafa8626:
All the fixes listed here are things myself and other users of the player found and they were tested a lot so that's why I decided to include them for this PR as patches, that eventually are going to be part of a release

westonruter:
@rafa8626 In that case, please update the `'4.2.5'` version references in `script-loader.php` to `'4.2.5-5d816f5'` or whatever the latest commit was when you copied ME.js into this PR. That would be helpful. You can see that this has been done for other scripts as well, using timestamps. But a commit hash is better (or even a timestamp _with_ a hash).

westonruter:
Who is going to be maintaining ME.js moving forward?

joemcgill:
> Who is going to be maintaining ME.js moving forward?

I'm curious to know the answer both from a dependency library point of view, and also on WP's side of things. Is this fix going back upstream to ME.js or is this specifically a WP back-compat thing?

rafa8626:
I'll do that Weston next week. And Joe, the changes are gonna be upstream on MEJs it's just that I don't know when I will release this given I'm not maintaining this project anymore. I don't know if John Dyer has found somebody but I'm checking PRs and integrating them occasionally

rafa8626:
@westonruter Done. I have updated that file with the commit hash

westonruter:
Committed to `trunk` in [r41320](https://core.trac.wordpress.org/changeset/41320).

PR #250, opened 2017-08-28
Change Close link to Done

timmyc:
This branch has the logic needed to change the "Close" link in the widgets to say "Done" instead

https://core.trac.wordpress.org/ticket/41610


src/wp-admin/includes/widgets.php
```diff
@@ -254,7 +254,7 @@ function wp_widget_control( $sidebar_args ) {
 	<div class="widget-control-actions">
 		<div class="alignleft">
 			<button type="button" class="button-link button-link-delete widget-control-remove"><?php _e( 'Delete' ); ?></button> |
-			<button type="button" class="button-link widget-control-close"><?php _e( 'Close' ); ?></button>
+			<button type="button" class="button-link widget-control-done"><?php _e( 'Done' ); ?></button>
```
westonruter: I think the class name should remain the same. I know some plugins hide this button entirely since it is redundant, and if the class name changes then they will break in this area.

So in that case the only change in this PR would be the text string change from `Close` to `Done`.

timmyc:
Great point @westonruter, we now have a 1-liner PR 😄

westonruter:
And the build passes! 😄

src/wp-admin/js/widgets.js
```diff
@@ -440,9 +440,15 @@ wpWidgets = {
...
 
 	save : function( widget, del, animate, order ) {
-		var self = this,
-			sidebarId = widget.closest('div.widgets-sortables').attr('id'),
-			data = widget.find('form').serialize(), a;
+		var self = this, data, a,
+			sidebarId = widget.closest( 'div.widgets-sortables' ).attr( 'id' ),
+			form = widget.find( 'form' );
+
+		if ( form.prop( 'checkValidity' ) && ! form[0].checkValidity() ) {
```
westonruter: Ideally this would also be implemented in the customizer, but the Customizer has a form around the entire document so widgets can't have individual forms.

westonruter:
Committed to trunk in r41352.

PR #251, opened 2017-08-31
Trac 39693: Fix missing assignment of widgets on theme switch

westonruter:
https://core.trac.wordpress.org/ticket/39693

### DIFF
--- a/src/wp-includes/widgets.php
+++ b/src/wp-includes/widgets.php
@@ -1107,107 +1107,189 @@ function _wp_sidebars_changed() {
  *
  * @param string|bool $theme_changed Whether the theme was changed as a boolean. A value
  *                                   of 'customize' defers updates for the Customizer.
- * @return array|void
+ * @return array
  */
 function retrieve_widgets( $theme_changed = false ) {
 	global $wp_registered_sidebars, $sidebars_widgets, $wp_registered_widgets;
 
-	$registered_sidebar_keys = array_keys( $wp_registered_sidebars );
-	$orphaned = 0;
+	$registered_sidebars_keys = array_keys( $wp_registered_sidebars );
+	$registered_widgets_ids   = array_keys( $wp_registered_widgets );
+	$old_sidebars_widgets     = get_theme_mod( 'sidebars_widgets' );
 
-	$old_sidebars_widgets = get_theme_mod( 'sidebars_widgets' );
 	if ( is_array( $old_sidebars_widgets ) ) {
 		// time() that sidebars were stored is in $old_sidebars_widgets['time']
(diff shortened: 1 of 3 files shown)


src/wp-includes/widgets.php
```diff
@@ -1107,107 +1107,191 @@ function _wp_sidebars_changed() {
...
+		} else {
+			$new_sidebars_widgets[ $sidebar ] = array();
 		}
+	}
 
-		$_sidebars_widgets[$sidebar] = $_widgets;
-		$shown_widgets = array_merge($shown_widgets, $_widgets);
+	// If there are no old sidebars left, then we're done.
+	if ( empty( $old_sidebars_widgets ) ) {
+		return $new_sidebars_widgets;
```
westonruter: I think this is returning prematurely.

src/wp-includes/widgets.php
```diff
@@ -1107,107 +1107,191 @@ function _wp_sidebars_changed() {
...
+
+/**
+ * Compares a list of sidebars with their widgets against a whitelist.
+ *
+ * @since 4.9.0
+ *
+ * @param array $old_sidebars_widgets List of sidebars and their widget instance IDs.
+ * @return array Mapped sidebars widgets.
+ */
+function _wp_map_sidebars( $old_sidebars_widgets ) {
```
westonruter: Might as well make this public since `wp_map_nav_menu_locations` is public.

westonruter: The `wp_map_nav_menu_locations` function takes two arguments: `$new_nav_menu_locations, $old_nav_menu_locations`. Perhaps this should take the `$new_sidebars_widgets` as well?

westonruter: Probably can rename this from `_wp_map_sidebars()` to `wp_map_sidebars_widgets` for better alignment with `wp_map_nav_menu_locations`.

src/wp-includes/widgets.php
```diff
@@ -1107,107 +1107,191 @@ function _wp_sidebars_changed() {
...
-					$_sidebars_widgets[$id] = $sidebars_widgets[$id];
-					unset( $sidebars_widgets[$id] );
-				}
-			}
+	// Find hidden/lost multi-widget instances.
+	$shown_widgets = call_user_func_array( 'array_merge', $sidebars_widgets );
+	$lost_widgets  = array_diff( $registered_widgets_ids, $shown_widgets );
+
+	foreach ( $lost_widgets as $key => $widget_id ) {
+		$number = preg_replace( '/.+?-([0-9]+)$/', '$1', $widget_id );
```
westonruter: This will blow away old single widgets.

PR #256, opened 2017-09-14
#39896: Customizer: Allow users to Draft changes before Publishing: Autosaving

westonruter:
https://core.trac.wordpress.org/ticket/39896

Todo:

- [x] Add filter for determining whether multiple concurrent changesets are allowed.
- [x] Prevent injecting changeset UUID into URL when in linear mode.
- [x] Ensure autosaved param is sent to preview immediately upon change and not just when the autosaved request finishes?
- [ ] Add post locking?
- [ ] Add unit tests.

Also closes wpcore#39275.


joyously:
Would it make sense for the autosave to have some sort of visual indicator for the fields being saved?
Or in general, each option would have an indicator for "changed", so you know what you are saving (or can tell whether you want to save).

westonruter:
Having an indicator next to each control for which there are dirty setting(s) has been mentioned as part of https://core.trac.wordpress.org/ticket/21666

I think that should be tackled as part of that ticket.

dlh01:
@westonruter Do you think any future `WP_Customize_Changeset` class (https://core.trac.wordpress.org/ticket/40527) would eventually be the source of logic for determining `$autosave_revision_post_id` and other autosave-related values?

westonruter:
@dlh01 that's a great question, and actually it's something I wanted to get your thoughts on. How exactly the `$autosaved` flag fits into the PHP side of things is very much just an initial stab. I welcome your thoughts on where best it will fit.

My initial thought on how it could fit is inside of `WP_Customize_Manager::changeset_data()`, where if `WP_Customize_Manager::$autosaved` is `true`, that this method could then check to see if there is an autosave revision for the `$changeset_post_id` for the current user (via `wp_get_post_autosave()`) and then if so, pass that revision ID into `WP_Customize_Manager::get_changeset_post_data()` to get back the content of revision instead of the content of the main changeset. The `get_changeset_post_data` method would need to be changed to explicitly allow a changeset's revision post ID to be passed in.

How exactly this plays into `WP_Customize_Changeset` is something you'd know better. Maybe there could be a `WP_Customize_Changeset::get_autosaved_data( $user_id )`?

src/wp-admin/js/customize-controls.js
```diff
@@ -5009,7 +5033,6 @@
 			api.bind( 'change', function() {
 				if ( state( 'saved' ).get() ) {
 					state( 'saved' ).set( false );
-					populateChangesetUuidParam( true );
```
westonruter: This needs to be retained when in non-linear changeset mode.

src/wp-admin/js/customize-controls.js
```diff
@@ -5022,6 +5045,7 @@
 				if ( 'publish' === response.changeset_status ) {
 					state( 'activated' ).set( true );
 				}
+				populateChangesetUuidParam( 'auto-draft' !== response.changeset_status );
```
westonruter: This should only be done in the default linear mode, which would be the default in core. When in non-linear mode, the UUID should get populated in the URL right away since multiple changesets can live concurrently.

src/wp-admin/js/customize-controls.js
```diff
@@ -5066,13 +5090,67 @@
 			};
 
 			changesetStatus.bind( function( newStatus ) {
-				populateChangesetUuidParam( '' !== newStatus && 'publish' !== newStatus );
+				populateChangesetUuidParam( '' !== newStatus && 'auto-draft' !== newStatus && 'publish' !== newStatus );
```
westonruter: The 'auto-draft' part of this is only applicable in the default linear mode.

src/wp-admin/js/customize-controls.js
```diff
@@ -5066,13 +5090,67 @@
...
+				// Remove parameter from the URL.
+				urlParser = document.createElement( 'a' );
+				urlParser.href = location.href;
+				queryParams = api.utils.parseQueryString( urlParser.search.substr( 1 ) );
+				delete queryParams.customize_autosaved;
+				urlParser.search = $.param( queryParams );
+				history.replaceState( {}, document.title, urlParser.href );
+			} else if ( api.settings.changeset.latestAutoDraftUuid || api.settings.changeset.hasAutosaveRevision ) {
+
+				// Since there is an autosave revision and the user hasn't loaded with autosaved, add notification to prompt to load autosaved version.
```
westonruter: This needs to be prevented if the Customizer was loaded with a changeset UUID provided.

src/wp-admin/js/customize-controls.js
westonruter: Actually, the changeset UUID really needn't be displayed at all in the default linear mode, since there is only ever one at a time.

westonruter: The UUID should not get stripped from the URL if it was initially supplied to begin with.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -2533,12 +2581,23 @@ function save_changeset_post( $args = array() ) {
 
 		// Note that updating a post with publish status will trigger WP_Customize_Manager::publish_changeset_values().
 		if ( $changeset_post_id ) {
-			$post_array['edit_date'] = true; // Prevent date clearing.
-			$r = wp_update_post( wp_slash( $post_array ), true );
+			if ( $args['autosave'] ) {
```
westonruter: Also needs if `'auto-draft' !== get_post_status( $changeset_post_id )`

src/wp-admin/js/customize-controls.js
westonruter: Eliminate dynamic injection of the `changeset_uuid` into the URL since it is not necessary for this linear mode of drafting. It still needs to be recognized for the sake of the auto-draft restore link, but as soon as that URL is followed the parameter can be stripped out of the location.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -2533,12 +2581,46 @@ function save_changeset_post( $args = array() ) {
...
+				}
+			}
 		} else {
 			$r = wp_insert_post( wp_slash( $post_array ), true );
 			if ( ! is_wp_error( $r ) ) {
 				$this->_changeset_post_id = $r; // Update cached post ID for the loaded changeset.
+
+				// @todo Limit this to linear mode?
+				// Delete all other auto-draft changeset posts for this user (they serve like autosave revisions), as there should only be one.
+				$autosave_autodraft_posts = get_posts( array(
```
westonruter: This call should get moved into a separate method to reduce duplication with the other query to get auto-drafts further below.

src/wp-admin/js/customize-controls.js
```diff
@@ -5436,26 +5518,70 @@
...
+			});
+
+			// Prompt user with AYS dialog if leaving the Customizer with unsaved changes
+			$( window ).on( 'beforeunload.customize-confirm', function() {
+				if ( ! isCleanState() ) {
+					setTimeout( function() {
+						overlay.removeClass( 'customize-loading' );
+					}, 1 );
+					return api.l10n.saveAlert;
+				}
```
sayedtaqui: Shouldn't this be returning `undefined` if condition is not true? `return undefined;`

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -2291,6 +2329,17 @@ function save_changeset_post( $args = array() ) {
 			$args['status'] = 'future';
 		}
 
+		// Validate autosave param. See _wp_post_revision_fields() for why these fields are disallowed.
+		if ( $args['autosave'] ) {
+			if ( $args['date_gmt'] ) {
+				return new WP_Error( 'illegal_autosave_with_date_gmt' );
+			} elseif ( $args['date_status'] ) {
```
dlh01: This line is generating a notice for me. Should it be `$args['status']`?

westonruter: Yes, you are right. Good thing I didn't use `! empty( … )` here!

dlh01:
I think there might be a potential collision between importing theme starter content and showing the "restore autosave" prompt under a certain scenario:

- Open the Customizer with a "fresh site" and a theme containing starter content.
- Make any change and verify that the change appears in the `auto-draft` changeset along with the starter content.
- Reload the Customizer.

Because there is no `changeset_uuid` in the URL, the Customizer will be bootstrapped with a new UUID. The theme starter content will be re-imported into a new changeset post under the new UUID, and the previous changeset will be deleted.

westonruter:
@dlh01 what is the conflict? If you click to restore a previous `auto-draft` changeset on a non-fresh site, then you'll be taken to the Customizer with the previous changeset and then the starter content will be loaded into that `auto-draft`. So is the re-importing the problem here? Or the discarding? I don't this would particularly be a new problem because if I open a theme in the Customizer normally, make a change and then reload, the starter content will get imported in that case as well.

dlh01:
> So is the re-importing the problem here? Or the discarding?

Both, in a way. In the scenario I was looking at, the re-importing causes the discarding. The changes I made before reloading are lost (because the `auto-draft` with the changes was deleted when importing the starter content into a new changeset), and there is no prompt to restore the autosave (because there is no more autosave).

> I don't this would particularly be a new problem because if I open a theme in the Customizer normally, make a change and then reload, the starter content will get imported in that case as well.

Except if the `changeset_uuid` is already in the URL, as I understand it. If the URL contains the `changeset_uuid`, reloading the Customizer would reload that UUID, including both the previously imported starter content and your change. 

However, it looked to me as if the `changeset_uuid` would no longer be automatically added to the URL with this patch, so reloading the Customizer always starts a changeset for a new UUID (given that there are no changesets in `array( 'draft', 'pending', 'future' )`.

westonruter:
@dlh01 I think I see the problem here. In particular, it's that a user's `auto-draft` changesets get deleted automatically when a new changeset is created for a user due to this logic:

https://github.com/xwp/wordpress-develop/blob/88ce81a8fb0e987e34398b0a3230dcb7ae32fd48/src/wp-includes/class-wp-customize-manager.php#L2606-L2624

So I think perhaps there is a combination of things that could be done to avoid this problem:

1. We could prevent the `auto-draft` garbage collection unless the changeset is being updated/inserted with a non-`auto-draft` status. The thinking there is that 
2. We could prevent the garbage collection unless you explicitly dismiss the notification in the same way this patch is [doing when you click the close button](https://github.com/xwp/wordpress-develop/blob/88ce81a8fb0e987e34398b0a3230dcb7ae32fd48/src/wp-admin/js/customize-controls.js#L5619-L5627).

But humm, maybe neither of these are the right solution. In fact, I've felt nervous about the auto-deletion of auto-drafts since up until now, this has been all that users have had, and we even told users they could copy the URL to the Customizer with the `changeset_uuid` intact to bookmark that session to come back to later (unless you wait longer than 7 days and the posts have been garbage collected).

Maybe instead of deleting the auto-drafts we should instead _mark_ the auto-drafts as having been “dismissed“. Instead of doing `wp_delete_post( $auto_draft_post_id )` we could instead `add_post_meta( $auto_draft_post_id, '_restore_dismissed', true )`. Then in the `get_posts()` call we can do a `meta_query` to exclude any auto-draft posts that have the `_restore_dismissed` meta key:

```php
'meta_query' => array( 
    array( 
        'key' => '_restore_dismissed',
        'compare' => 'NOT EXISTS' ,
    ),
),
```

The thing left to nail down then would be exactly when the dismissal should happen. 

1. If I open the Customizer on a fresh site, load the starter content, and make a bunch of changes.
2. I close the Customizer without saving anything.
3. I come back to the Customizer while the site is still fresh and starter content gets re-imported into a changeset. 
4. At this point I should get presented with the autosave prompt.
5. If I click the link to restore the autosave, then at that point all of the auto-draft changesets should get updated to have `_restore_dismissed` so that when I load the restored changeset, I don't get a notification to restore the _previous_ auto-draft I was just working on.
6. Alternatively, the moment that I decide to Save a changeset as a draft, scheduled, or published (https://github.com/xwp/wordpress-develop/pull/258) then at this point as well all other auto-draft changesets should be marked as dismissed.

Does this account for the scenarios?

dlh01:
I agree with you on the principle of not deleting auto-drafts. The previous advertising about the ability to copy the URL seems like reason enough to me to keep the behavior. Personally, I've also found it generally handy. 

I would want to think more to formulate an opinion about the `_restore_dismissed` approach specifically as the means for keeping it. The steps you described do sound like they would cover the scenario I outlined, though.

A related idea I was kicking around was to save the starter content changeset as a `draft` so subsequent changes to it are automatically stored as a revision. I haven't been able to work out that idea to completion, though.

westonruter:
@dlh01 @PatelUtkarsh @sayedtaqui @valendesigns @miina I would greatly appreciate your reviews on the changes here!

The changes here will have a key impact on Customize Snapshots. In particular, the plugin will need to add:

```php
add_filter( 'customize_changeset_branching', '__return_true' );
```

This essentially restores the 4.7 behavior in core for how changesets are concurrent, as you can see in the filter docs in this PR. You can add that filter yourself in an mu-plugin to see the behavior then to be largely the same as 4.7-4.8, with the added improvement of there being autosave detection. If this filter is not added, where branching is false by default, then if you do `wp.customize.previewer.save({status: 'draft'})` then when you reload the Customizer you'll automatically be restored to that changeset without starting a new one from scratch.

src/wp-includes/theme.php
```diff
@@ -2796,19 +2796,70 @@ function _wp_customize_include() {
...
+	 *
+	 * Since linear changesets are deemed to be more suitable for the majority of WordPress users,
+	 * they are the default. For WordPress sites that have heavy site management in the Customizer
+	 * by multiple users then branching changesets should be enabled by means of this filter.
+	 *
+	 * @since 4.9.0
+	 * @param bool $allow_branching Whether branching is allowed. If `false`, the default,
+	 *                              then only one saved changeset exists at a time.
+	 */
+	$branching = apply_filters( 'customize_changeset_branching', false );
```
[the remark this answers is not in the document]
westonruter: What situations should it be filterable from a theme? And yes, since the `setup_theme` action is used that would be a reason for why a theme may be too late.

Really the filter needs to apply early enough that it can effect which UUID is fed into `WP_Customize_Manager`. If branching is not allowed, then the UUID would come from the most recent drafted/scheduled changeset. Otherwise, if branching is allowed then a new UUID should be generated.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -2519,12 +2621,35 @@ function save_changeset_post( $args = array() ) {
...
 		} else {
 			$r = wp_insert_post( wp_slash( $post_array ), true );
 			if ( ! is_wp_error( $r ) ) {
 				$this->_changeset_post_id = $r; // Update cached post ID for the loaded changeset.
+
+				// @todo Limit this to linear mode?
+				// Delete all other auto-draft changeset posts for this user (they serve like autosave revisions), as there should only be one.
+				foreach ( $this->get_autodraft_changesets() as $autosave_autodraft_post ) {
+					if ( $autosave_autodraft_post->ID !== $this->_changeset_post_id ) {
+						update_post_meta( $autosave_autodraft_post->ID, '_customize_restore_dismissed', true );
```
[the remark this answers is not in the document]
westonruter: Yes, good point. In that case, the autosave needs to be retrieved earlier in `WP_Customize_Manager` before we end up calling `_save_starter_content_changeset`, that is, at the `wp_loaded`/`customize_register` action. In the `setup_theme` action?  It would need to happen before `after_theme_setup` that's for sure, which is way before  `WP_Customize_Manager::customize_pane_settings()` where it currently occurs. What about the constructor? That doesn't feel great either.

westonruter:
> Apart from the two minor comments inline, I wanted to confirm one part of the expected behavior:

@dlh01 that is correct.

src/wp-includes/class-wp-customize-manager.php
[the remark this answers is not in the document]
westonruter: Or maybe the dismissal should happen in a different way, with an explicit request?

src/wp-admin/js/customize-controls.js
westonruter: It has been retained, but to chun below.

src/wp-includes/theme.php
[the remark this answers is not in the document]
dlh01: Just in some client or hosting situations where we work primarily in the theme, and access to `plugins` or `mu-plugins` is difficult to come by. In those cases, we would want to enable branching from within the theme because that's where our other Customizer-specific logic would be. But the problem has come up before, e.g. https://core.trac.wordpress.org/ticket/24169, so branching wouldn't be introducing anything new.

src/wp-includes/script-loader.php
```diff
@@ -561,6 +561,8 @@ function wp_default_scripts( &$scripts ) {
 		'expandSidebar'      => _x( 'Show Controls', 'label for hide controls button without length constraints' ),
 		'untitledBlogName'   => __( '(Untitled)' ),
 		'serverSaveError'    => __( 'Failed connecting to the server. Please try saving again.' ),
+		/* translators: placeholder is URL to the Customizer to load the autosaved version */
+		'autosaveNotice'     => __( 'There is a more recent autosave of your changes than the one you are previewing. <a href="%s">Restore the autosave</a>' ),
```
dlh01: When starting a new `auto-draft` session in the Customizer and this prompt is triggered by a previous `auto-draft` session, I don't think it's true that the autosave is "more recent," is it?

sayedtaqui:
@westonruter Tested with snapshot plugin, not sure why its not creating revisions even though the revision support is added for customize_changeset. 
https://github.com/xwp/wp-customize-snapshots/pull/161

sayedtaqui:
@westonruter In linear mode -

1 - If a user makes some changes leave customizer and comes back he is prompted with an option to restore the changes. 
2- If he saves the status as draft makes additional changes, leave and come back, he is again prompted to restore the changes.
3- If he saves the status as future, makes additional changes leave and come back to customizer before the scheduled time he is promoted with restore option.
4- However if he he comes back after the scheduled time, the status becomes publish which is correct, but he is not prompted to restore the changes.

a - Though we have the changes made in point-3 stored in a revision of the original trashed changeset, we would not prompt to restore those additional changes, is that considered lost? We won't follow the pattern used in point 1 and 2, 3?

b - And if that is intended, what if they are sitting on customizer and the scheduled time has arrived? I see you have added `$response['next_changeset_uuid'] = wp_generate_uuid4();` I think for this reason , so the new UUID will now be `next_changeset_uuid` and we will save those additional changes on that in `auto-draft` status and treat it as new changeset, correct?

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -2519,12 +2621,35 @@ function save_changeset_post( $args = array() ) {
...
+				}
+			}
 		} else {
 			$r = wp_insert_post( wp_slash( $post_array ), true );
 			if ( ! is_wp_error( $r ) ) {
 				$this->_changeset_post_id = $r; // Update cached post ID for the loaded changeset.
+
+				// @todo Limit this to linear mode?
+				// Delete all other auto-draft changeset posts for this user (they serve like autosave revisions), as there should only be one.
+				foreach ( $this->get_autodraft_changesets() as $autosave_autodraft_post ) {
```
sayedtaqui: It wouldn't be a problem in the default linear mode,  but `$this->get_autodraft_changesets()` would be okay for non-linear mode?, because there may be a lot of `auto-draft` changesets created, and the default `posts_per_page` for `get_autodraft_changesets` is `-1`.

sayedtaqui: or may be not, because how many can someone have in a week.

src/wp-includes/class-wp-customize-manager.php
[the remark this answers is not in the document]
dlh01: I agree that retrieving it in the constructor would feel awkward. I guess my first, tentative thought would be to retrieve it in the `setup_theme` callback, perhaps after the UUID is validated? In that case, the flow would read something like "verify the UUID and gather related information about it."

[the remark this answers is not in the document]
westonruter: @dlh01 unfortunately the `setup_theme` callback wouldn't help because it runs _before_ the `functions.php` is loaded for a theme. We'd have to defer to `after_setup_theme` priority `0`. Maybe that would work. It's just `import_theme_starter_content` happens there at priority `100` and I'd need to double-check that the Customizer doesn't try to read the UUID ever before that point.

src/wp-includes/theme.php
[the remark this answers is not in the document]
westonruter: You're right, and that is something we'd face on WordPress.com VIP non-Go hosting.

We might be able to defer to `after_setup_theme` priority `0`. Maybe that would work. It's just `import_theme_starter_content` happens there at priority `100` and I'd need to double-check that the Customizer doesn't try to read the UUID ever before that point.

If we can't get this to work, then the other option that comes to mind is to introduce support for a new option that can be used instead:

```diff
--- src/wp-includes/theme.php
+++ src/wp-includes/theme.php
@@ -2827,7 +2827,7 @@ function _wp_customize_include() {
 	 * @param bool $allow_branching Whether branching is allowed. If `false`, the default,
 	 *                              then only one saved changeset exists at a time.
 	 */
-	$branching = apply_filters( 'customize_changeset_branching', false );
+	$branching = apply_filters( 'customize_changeset_branching', (bool) get_option( 'customize_changeset_branching', false ) );
 
 	// Automatically fetch the most recent drafted changeset when changeset branching is not enabled.
 	if ( ! $branching && empty( $changeset_uuid ) ) {
```

That would then allow a theme to set the option the first time it loads and then it would be set thereafter.

src/wp-includes/class-wp-customize-manager.php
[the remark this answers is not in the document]
westonruter: Sorry, that comment was meant as a response to your other question: https://github.com/xwp/wordpress-develop/pull/256#discussion_r140851791

But as I commented there, I think we could indeed defer resolving the UUID until the first time we actually need it. Maybe that could be `after_setup_theme`, as long as it is before importing the starter content which happens at priority 100.

westonruter:
> Tested with snapshot plugin, not sure why its not creating revisions even though the revision support is added for customize_changeset.

I see why. It's because `customize_autosaved` is getting included in the request, which in turn results in this short-circuit:

https://github.com/xwp/wordpress-develop/blob/d6a58eec48261394477683b727c14b7390d56465/src/wp-includes/revision.php#L112-L114

src/wp-includes/script-loader.php
westonruter: I think it is true, because the when you load the Customizer anew and there are only `auto-draft` changesets, then you load the Customizer in a sate where it shows settings that were published _before_ any of the other `auto-draft` changesets were created.

src/wp-includes/class-wp-customize-manager.php
westonruter: Good question. But the auto-drafts returned here are all tied to the current user, and only the ones that have not been dismissed yet. I don't see how this could possibly result in a problematic number of auto-draft changesets being returned, since they get garbage collected after 7 days, and they'll also get excluded from the next call to `get_autodraft_changesets()` since `$exclude_restore_dismissed` is true by default. So every time they open the Customizer they'd be re-dismissing the most recent auto-drafts.

westonruter:
@sayedtaqui you raise some interesting scenarios in https://github.com/xwp/wordpress-develop/pull/256#issuecomment-331853262

> However if he he comes back after the scheduled time, the status becomes publish which is correct, but he is not prompted to restore the changes.
> a - Though we have the changes made in point-3 stored in a revision of the original trashed changeset, we would not prompt to restore those additional changes, is that considered lost? We won't follow the pattern used in point 1 and 2, 3?

This is a good point. Either I think it is fine to not prompt them to restore restore changes made after scheduling a changeset, or it would be sufficiently rare enough to not be worth the time to address right now. If we did want to address it, then I could see that if there is an autosave revision for a scheduled changeset then it could get copied into a _new_ auto-draft changeset to prompt the user to restore when coming back to the Customizer. I don't foresee this happening often to users, however, and if someone schedules a changeset and then leaves without saving those additional changes, then I think a user should expect those changes to have been abandoned and not prompt to restore them. @melchoyce what do you think?

src/wp-includes/script-loader.php
dlh01: Ah, I see what you mean.

src/wp-includes/class-wp-customize-manager.php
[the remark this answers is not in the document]
westonruter: @dlh01 I've decided to just opt-out of providing a user with autosave restoration in the case where starter content had been loaded. Since it's a fresh site after all, it's unlikely that there will be anything to restore. The starter content is in essence a static auto-save, so prompting to restore an autosave on top of starter content would be confusing.

src/wp-includes/theme.php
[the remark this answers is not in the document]
westonruter: I've implemented this. There is now a `WP_Customize_Manager::establish_loaded_changeset()` method which is called at `after_setup_theme` to both apply the `customize_changeset_branching` filter and also to select either an existing saved changeset or generate a random one. I've done it in a way where it should be backwards compatible.

src/wp-includes/class-wp-customize-manager.php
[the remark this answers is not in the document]
dlh01: That sounds sensible to me.

src/wp-includes/class-wp-customize-manager.php
sayedtaqui: You are right, that's what I thought after leaving the first comment.

westonruter:
Committed to trunk in r41597.

westonruter:
Remaining work for this feature is being done in https://github.com/xwp/wordpress-develop/pull/263

melchoyce:
> if someone schedules a changeset and then leaves without saving those additional changes, then I think a user should expect those changes to have been abandoned and not prompt to restore them. @melchoyce what do you think?

Agreed

PR #257, opened 2017-09-16
Trac 41897: Introduce reusable/extendable WP_Customize_Code_Editor_Control and use for Custom CSS

westonruter:
See https://core.trac.wordpress.org/ticket/41897


westonruter:
Committed to `trunk` in r41558.

PR #258, opened 2017-09-17
Trac #39896-UI: Customizer: Allow users to Draft changes before Publishing

sayedtaqui:
https://core.trac.wordpress.org/ticket/39896

**Todo Tasks**
- ~Slide up / slide down animation of the publish settings section,~
- [x] Switch publish settings to be positioned and animated in the same way as available widgets and nav menu panels.
- [x] Save and gear button behaviour.
- [x] Share Preview link and make the link clickable.
- [x] Make the schedule date control working.
- [x] When publish settings are open I think the cog half of the publish button should be "pressed" (via @melchoyce)
- [x] Improve `wp.customize.previewer.save()` for changesets. See [#39275](https://core.trac.wordpress.org/ticket/39275)
- [x] Auto publish if sitting on Customizer with changeset scheduled and time arrived
- [x] De-activate the themes section when a changeset is not autodraft.
- [x] Add unit tests for code added in #256.
- [ ] Add post locking?


src/wp-admin/js/customize-controls.js
```diff
@@ -5838,6 +5857,36 @@
...
+		api.control( 'changeset_preview_link', function( control ) {
+			var copyButton, previewLink;
+
+			copyButton = $( wp.template( 'customize-copy-preview-link' )() );
+
+			copyButton.on( 'click', function( event ) {
+				event.preventDefault();
+				previewLink = control.container.find( 'input' );
+
+				if ( previewLink.val() ) {
```
sayedtaqui: @westonruter  This isn't looking good, let's get https://core.trac.wordpress.org/ticket/37964 
first.

src/wp-admin/js/customize-controls.js
```diff
@@ -4527,8 +4527,27 @@
...
 			saveBtn = $( '#save' ),
+			publishSettingsBtn = $( '#publish-settings' ),
 			footerActions = $( '#customize-footer-actions' );
 
+		api.section( 'publish_settings', function( section ) {
+			publishSettingsBtn.prop( 'disabled', false );
+			publishSettingsBtn.on( 'click', function( event ) {
+				event.preventDefault();
+
+				// @todo Change expand animation to SlidUp/SlideDown.
```
westonruter: It may be a CSS thing instead of a `slideUp`/`slideDown`. I recall CSS transitions are being used.

src/wp-admin/js/customize-controls.js
westonruter: I didn't anticipate the preview link as being directly a concern for #37964. How would it be used for the preview link? Or are you referring to the status radio buttons, and they aren't implemented yet here because they won't look good?

src/wp-admin/js/customize-controls.js
```diff
@@ -4527,8 +4527,27 @@
...
+				if ( section.expanded.get() ) {
+					section.collapse();
```
westonruter: I just realized an unexpected side-effect of using a section for storing the publish setting controls. If I down editing a widget control with the sidebar section expanded, and then if I hit the gear icon to edit the publish settings, it will collapse the sidebar section and widgets panel to then expand the publish settings section. Currently if I collapse the publish settings section, then I am left at the Customizer root and I'd have to navigate my way back up to where I was. At the very least the panel/section/control that was expanded should be re-focused when the section is collapsed. But maybe we're trying too much to hack a Section for our needs here and we should instead create the publish settings as a different kind of panel entirely? That would make it difficult for plugins to add their own controls to the publish settings area, however. Not sure if that should be allowed presently.

One other thing that comes to mind is perhaps expanding the publish settings section could have an exception in place where all other sections do not get collapsed first. In particular, it would be about disabling this bit of code:

https://github.com/xwp/wordpress-develop/blob/2d1ac710e270f2c70d26a685c014efd3214151df/src/wp-admin/js/customize-controls.js#L1241-L1245

There's that `allowMultiple` in there which is some dead code carried over from when sections were not panels but just accordions and so it allowed you to drag and drop widgets between sections. Maybe that could be resurrected for our needs here, except it would need to make sure that the section you are expanding gets the highest `z-index` to make sure it always appears at the top.

src/wp-admin/js/customize-controls.js
westonruter: I played around with it in https://github.com/xwp/wordpress-develop/pull/258/commits/50b9d22f43ec37d658361a75e086e6652271ed5f and I don't think that #37964 is necessarily a dependency. It's a little code to link a setting-less control to a `wp.customize.Value`.

Demo: https://youtu.be/Q2-n1LCOnbg

src/wp-admin/js/customize-controls.js
```diff
@@ -4958,6 +4977,7 @@
 				expandedPanel = state.create( 'expandedPanel' ),
 				expandedSection = state.create( 'expandedSection' ),
 				changesetStatus = state.create( 'changesetStatus' ),
+				nextChangesetStatus = state.create( 'nextChangesetStatus' ),
```
westonruter: I'm not sure that `nextChangesetStatus` is the best name. Maybe `selectedChangesetStatus` would be better.

src/wp-admin/js/customize-controls.js
```diff
@@ -5838,6 +5881,68 @@
...
+		api.control( 'changeset_preview_link', function( control ) {
+			control.deferred.embedded.done( function() {
+				var copyButton, inputNode, element, getLink;
+
+				copyButton = $( wp.template( 'customize-copy-preview-link' )() );
+				inputNode = control.container.find( 'input' );
+				element = new api.Element( inputNode );
+				control.elements.push( element );
+
+				// @todo Need to improve and test it.
```
westonruter: I think instead of an `input` it should actually be an `a[href][target=_blank]` so that they can just click on it to open it in a new window. 

@melchoyce Maybe the URL shouldn't even be shown. They can only see the domain part of it anyway:

image

Why not just have a button for Frontend Preview and another to Copy the URL?

src/wp-admin/js/customize-controls.js
```diff
@@ -5838,6 +5881,68 @@
...
+				getLink = function() {
+					var a = document.createElement( 'a' ), params = {};
+
+					params.customize_changeset_uuid = api.settings.changeset.uuid;
+					a.href = api.previewer.previewUrl.get();
+
+					if ( ! api.settings.theme.active ) {
+						params.theme = api.settings.theme.stylesheet;
```
westonruter: This isn't necessary because the publish settings won't be displayed when the theme is not active.

src/wp-admin/js/customize-controls.js
```diff
@@ -5838,6 +5881,68 @@
...
+
+					return a.href;
+				};
+
+				element.set( getLink() );
+
+				copyButton.on( 'click', function( event ) {
+					event.preventDefault();
+
+					if ( element.get() ) {
```
westonruter: I thin this function should incorporate @miina's work in https://github.com/xwp/wp-customize-snapshots/pull/116 to make sure that whether or not the user clicks to copy the URL or they click to open the URL in a new tab, that it makes sure that the changeset has been updated before proceeding.

Should clicking the URL or the button cause the equivalent of clicking the Save button? In other words, should any autosave revision changes be saved automatically?

src/wp-admin/js/customize-controls.js
```diff
@@ -5838,6 +5881,68 @@
...
+				} );
+
+				copyButton.on( 'mouseenter', function() {
+					if ( element.get() ) {
+						copyButton.focus();
+						copyButton.text( copyButton.data( 'copy-text' ) );
+					}
+				} );
+
+				control.container.append( copyButton );
```
westonruter: Maybe this would warrant a custom control template rather than amending custom elements into a default control?

src/wp-admin/css/customize-controls.css
```diff
@@ -53,6 +75,78 @@ body {
...
+
+#customize-control-changeset_preview_link input {
+	line-height: 2.5;
+	border-top: 1px solid #ddd;
+	border-left: none;
+	border-right: none;
+}
+
+#accordion-section-publish_settings {
+	display: none!important;
```
westonruter: Missing a space before `!`

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -3224,6 +3227,24 @@ public function render_control_templates() {
...
+	 * @since 4.9.0
+	 */
+	public function render_publish_settings_templates() {
+		?>
+		<script type="text/html" id="tmpl-customize-schedule-date">
+			<div id="customize-schedule-date" class="customize-schedule-date"></div>
+		</script>
+
+		<script type="text/html" id="tmpl-customize-copy-preview-link">
+			<button id="customize-copy-preview-link" class="customize-copy-preview-link button button-secondary" data-copy-text="<?php _e( 'Copy' ); ?>" data-copied-text="<?php _e( 'Copied' ); ?>" ><?php _e( 'Copy' ); ?></button>
```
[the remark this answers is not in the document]
sayedtaqui: I was trying to follow the pattern used in core, at most places just `_e()` is used not `esc_html_e()` for translated strings, so did not use `esc_attr_e()` but will do 👍

src/wp-admin/js/customize-controls.js
sayedtaqui: Agree, I think an additional frontend preview button would be helpful

src/wp-admin/js/customize-controls.js
```diff
@@ -5838,6 +5881,68 @@
...
+				copyButton = $( wp.template( 'customize-copy-preview-link' )() );
+				inputNode = control.container.find( 'input' );
+				element = new api.Element( inputNode );
+				control.elements.push( element );
+
+				// @todo Need to improve and test it.
+				getLink = function() {
+					var a = document.createElement( 'a' ), params = {};
+
+					params.customize_changeset_uuid = api.settings.changeset.uuid;
```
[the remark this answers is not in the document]
sayedtaqui: This is a todo task, i just wanted to get a placeholder link to work with the UI.

src/wp-admin/customize.php
```diff
@@ -133,13 +133,13 @@
...
-			if ( ! current_user_can( get_post_type_object( 'customize_changeset' )->cap->publish_posts ) ) {
-				$save_attrs['style'] = 'display: none';
-			}
-			submit_button( $save_text, 'primary save', 'save', false, $save_attrs );
+			$save_text = $wp_customize->is_theme_active() ? __( 'Publish' ) : __( 'Activate &amp; Publish' );
+			$capability = current_user_can( get_post_type_object( 'customize_changeset' )->cap->publish_posts );
 			?>
+			<div id="customize-save-button-wrapper" class="customize-save-button-wrapper <?php echo $capability ? '' : 'hidden'; ?>" >
+				<?php submit_button( $save_text, 'primary save', 'save', false ); ?>
+				<button id="publish-settings" class="publish-settings button-primary button dashicons dashicons-admin-generic" aria-label="<?php _e( 'Publish Settings' ); ?>" aria-expanded="false" disabled></button>
```
westonruter: Needs `esc_attr_e()` in attributes.

src/wp-admin/js/customize-controls.js
```diff
@@ -4974,21 +5081,36 @@
...
 				 */
-				canSave = ! saving() && ( ! activated() || ! saved() || ( '' !== changesetStatus() && 'publish' !== changesetStatus() ) );
+				canSave = ! saving() && ( ! activated() || ! saved() || ( changesetStatus() !== selectedChangesetStatus() && '' !== changesetStatus() ) );
 
 				saveBtn.prop( 'disabled', ! canSave );
 			});
 
+			selectedChangesetStatus.validate = function( status ) {
+				if ( '' === status || 'auto-draft' === status ) {
+					return 'publish';
```
westonruter: This seems to be causing a problem if I:

1. Make a change on a fresh customizer state.
2. Immediately go to Publish Settings and select Draft
3. Click in the preview to trigger an autosave.

The result is that the status gets changes from Draft back to Publish. It seems that maybe this `validate` method should return `null` here instead to prevent and change to the `selectedStatus`. This logic was originally added to reset the status to be `publish` after publishing changes, but that shouldn't be needed normally unless someone schedules a change that is not in the future.

src/wp-includes/customize/class-wp-customize-date-time-control.php
```diff
@@ -0,0 +1,169 @@
...
+						if ( 0 === $current_offset ) {
+							$tz_string = 'UTC+0';
+						} elseif ( $current_offset < 0 ) {
+							$tz_string = 'UTC' . $current_offset;
+						} else {
+							$tz_string = 'UTC+' . $current_offset;
+						}
+					}
+					?>
+					<span class="date-timezone"><?php echo esc_html( $tz_string ); ?></span>
```
westonruter: Maybe we should make this an `abbr` with the full timezone name being in the `title`? This would be similar to https://github.com/xwp/wp-customize-posts/blob/95505664068cab4b89b13a37e8bcdba00cbbc57f/php/class-wp-customize-post-date-control.php#L71-L101

westonruter: I copied the above logic from core, but I perhaps should have copied what we wrote in Customize Posts instead.

src/wp-admin/js/customize-controls.js
sayedtaqui: @westonruter Not just sections, the panels are also being collapsed right below this code https://github.com/xwp/wordpress-develop/blob/2d1ac710e270f2c70d26a685c014efd3214151df/src/wp-admin/js/customize-controls.js#L1254-L1256

So I am able to hack and not let the sections collapse but if we are on a panel which has sections, it will come back to root.
( The way I tried to hack the collapse )
```
section.expandedArgumentsQueue = [ {
	unchanged: true,
	allowMultiple: true,
	completeCallback: function() {
               // Where active class will have the animation
		section.container.toggleClass( 'active', section.expanded.get() );
	}
} ];

section.expanded.set( ! section.expanded.get() );
``
Trying to figure out a way..

File: src/wp-admin/js/customize-controls.js [hunk 15]
```diff
@@ -4620,7 +4726,7 @@
 			save: function( args ) {
 				var previewer = this,
 					deferred = $.Deferred(),
-					changesetStatus = 'publish',
+					changesetStatus = api.state( 'selectedChangesetStatus' ).get(),
```
westonruter: @sayedtaqui I'm thinking that there could be a `changesetDate` added here as well, and then it could perhaps get its date from a `changesetDate` state value, unless an `args.date` is supplied? And only supply the date if `'future' === changesetDate`.

src/wp-admin/js/customize-controls.js
westonruter: You're right, any open panels and sections get collapsed. If `allowMultiple === true` maybe it could skip the `panel.collapse()` part?

sayedtaqui: So wrap the below code in `allowMultiple !== true` condition?

src/wp-admin/js/customize-controls.js
```diff
@@ -1251,9 +1251,11 @@
 						completeCallback: expand
 					});
 				} else {
-					api.panel.each( function( panel ) {
-						panel.collapse();
-					});
+					if ( ! args.allowMultiple ) {
```
sayedtaqui: @westonruter Please see if adding condition here would be fine. Its now working the way we want.

src/wp-admin/css/customize-controls.css
```diff
@@ -165,6 +165,25 @@ body {
...
 }
 
+#customize-theme-controls .accordion-section-content#sub-accordion-section-publish_settings {
+	transform: translate( 0, -100% );
+	left: 0;
+	top: 0;
+}
+
+#sub-accordion-section-publish_settings .customize-section-back {
+	display: none;
```
sayedtaqui: Need to add the arrow up and better animation.

src/wp-admin/js/customize-controls.js
westonruter: @sayedtaqui Looks good to me. 👍 You're resurrecting `allowMultiple` from oblivion.

src/wp-includes/customize/class-wp-customize-date-time-control.php
sayedtaqui: Correct, I will get it from customize-posts.

src/wp-admin/js/customize-controls.js
```diff
@@ -3629,6 +3631,91 @@
...
+				if ( element.get() ) {
+					inputNode.select();
+					document.execCommand( 'copy' );
+					copyButton.text( copyButton.data( 'copied-text' ) );
+				}
+			} );
+
+			copyButton.on( 'mouseenter', function() {
+				if ( element.get() ) {
+					copyButton.focus();
```
westonruter: Why focus on `mouseenter`?

sayedtaqui: When you click on copy button, it selects the link from the input and the button text changes to "copied". Now if you want to copy again, the text should become `copy` and link in the input should be deselected. 
So I was thinking that when to change the text back to `copy`. Currently I am doing it when you hover over the copy button. `copyButton.focus()` will deselect the link in the input. There is also `.blur()` but not sure why its not working in this case.

westonruter: It just seems a little strange for the Copy button to get focus as soon as I hover over it.

src/wp-admin/customize.php
```diff
@@ -133,13 +133,13 @@
...
-			$save_text = $wp_customize->is_theme_active() ? __( 'Save &amp; Publish' ) : __( 'Save &amp; Activate' );
-			$save_attrs = array();
-			if ( ! current_user_can( get_post_type_object( 'customize_changeset' )->cap->publish_posts ) ) {
-				$save_attrs['style'] = 'display: none';
-			}
-			submit_button( $save_text, 'primary save', 'save', false, $save_attrs );
+			$save_text = $wp_customize->is_theme_active() ? __( 'Publish' ) : __( 'Activate &amp; Publish' );
+			$capability = current_user_can( get_post_type_object( 'customize_changeset' )->cap->publish_posts );
 			?>
+			<div id="customize-save-button-wrapper" class="customize-save-button-wrapper <?php echo $capability ? '' : 'hidden'; ?>" >
```
westonruter: I don't think the `$capability` is being used right here. In particular, if a user cannot `publish_posts` they may still be able to create a draft. In that case, the initial `selectedChangsetStatus` should be `draft` instead of `publish`, and `publish` can just be removed from the `choices`.

src/wp-admin/js/customize-controls.js
sayedtaqui: Thanks for pointing out , will try to figure out a better way.

src/wp-admin/css/customize-controls.css
```diff
@@ -53,10 +75,191 @@ body {
...
+	border-right-color: #0073aa;
+	outline: none;
+	box-shadow: none;
+}
+
+#customize-theme-controls .accordion-section-content#sub-accordion-section-publish_settings {
+	transform: translate( 0, -100% );
+	left: 0;
+	top: 0;
+	z-index: 99999;
```
westonruter: So this is how the `allowMultiple` section is forced to be displayed over other sections and panels, right?

I wonder if this could be generalized so that instead of the `z-index` being defined in CSS, if there was a variable `zIndex = 99999` in JS, and every time that you open a panel or section with `allowMultiple` that it would increment that `zIndex` and assign it to the container. In that way, you could create a stack of sections on top of each other `allowMultiple`. This could be quite a nice improvement elsewhere in core and plugins where we have the `focusConstructWithBreadcrumb` thing in core and Customize Posts:

https://github.com/xwp/wordpress-develop/blob/e7bbb482d6069d9c2d0e33789c7d290ac231f056/src/wp-admin/js/customize-widgets.js#L2176-L2193

https://github.com/xwp/wp-customize-posts/blob/f7b62df6744eb3eb54eb1aec9d877d51ee94cfff/js/customize-posts.js#L892-L931

Currently it collapses a given section before expanding and focusing on the other one, and then when that section gets collapsed it then returns to expand the other one. When jumping between a nav menu widget in the widgets panel to a menu section in the menus panel, the amount of panels moving around can be intense, and the user won't necessarily be brought back to the same scroll position they left on.

westonruter: When such an `allowMultiple` section is collapsed, then the inline `z-index` would be emptied to restore it's original value from the stylesheet.

westonruter:
So having a `z-index: 99999` in the stylesheet is how the `allowMultiple` section is forced to be displayed over other sections and panels, right?

I wonder if this could be generalized so that instead of the `z-index` being defined in CSS, if there was a variable `zIndex = 99999` in JS, and every time that you open a panel or section with `allowMultiple` that it would increment that `zIndex` and assign it to the container. In that way, you could create a stack of sections on top of each other `allowMultiple`. This could be quite a nice improvement elsewhere in core and plugins where we have the `focusConstructWithBreadcrumb` thing in core and Customize Posts:

https://github.com/xwp/wordpress-develop/blob/e7bbb482d6069d9c2d0e33789c7d290ac231f056/src/wp-admin/js/customize-widgets.js#L2176-L2193

https://github.com/xwp/wp-customize-posts/blob/f7b62df6744eb3eb54eb1aec9d877d51ee94cfff/js/customize-posts.js#L892-L931

Currently it collapses a given section before expanding and focusing on the other one, and then when that section gets collapsed it then returns to expand the other one. When jumping between a nav menu widget in the widgets panel to a menu section in the menus panel, the amount of panels moving around can be intense, and the user won't necessarily be brought back to the same scroll position they left on.

When such an `allowMultiple` section is collapsed, then the inline `z-index` would be emptied to restore it's original value from the stylesheet.

src/wp-includes/customize/class-wp-customize-date-time-control.php
```diff
@@ -112,7 +99,7 @@ public function content_template() {
 							<option value="pm"><?php esc_html_e( 'PM' ); ?></option>
 						</select>
 					</label>
-					<span class="date-timezone" aria-label="<?php esc_attr_e( 'Timezone' ); ?>" title="<?php esc_attr_e( $timezone_info['description'] ) ?>"><?php esc_html_e( $timezone_info['abbr'] ); ?></span>
+					<span class="date-timezone" aria-label="<?php esc_attr_e( 'Timezone' ); ?>" title="<?php echo esc_attr( $timezone_info['description'] ); ?>"><?php echo esc_html( $timezone_info['abbr'] ); ?></span>
```
sayedtaqui: 😬

src/wp-includes/customize/class-wp-customize-date-time-control.php
```diff
@@ -0,0 +1,238 @@
...
+					<label class="hour-field">
+						<span class="screen-reader-text"><?php esc_html_e( 'Hour' ); ?></span>
+						<input type="number" size="2" maxlength="2" autocomplete="off" class="date-input hour" data-component="hour" min="1" max="12" value="{{ data.hour }}" />
+					</label>
+					<span class="time-special-char date-time-separator">:</span>
+					<label class="minute-field">
+						<span class="screen-reader-text"><?php esc_html_e( 'Minute' ); ?></span>
+						<input type="number" size="2" maxlength="2" autocomplete="off" class="date-input minute" data-component="minute" min="0" max="59" value="{{ data.minute }}" />
+					</label>
+					<label class="am-pm-field">
```
westonruter: This isn't currently taking into account the `saveTwelveHourFormat`, but should we have it?

The core post editor does not have AM/PM:

image

@melchoyce should it be different here?

melchoyce: Yeah, I think the core post editor is confusing and I find the use of 24 hour time hella confusing for most Americans — was hoping to make it less confusing here, which is why I added am/pm. Could be good to make this an enhancement, where you only see AM/PM if you're in a timezone that uses it.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -3736,6 +3739,52 @@ public function register_controls() {
...
+			'capability' => 'customize',
+		) );
+
+		$this->add_control( new WP_Customize_Date_Time_Control( $this, 'changeset_scheduled_date', array(
+			'section' => 'publish_settings',
+			'settings' => array(),
+			'type' => 'date_time',
+			'min_year' => date( 'Y' ),
+			'allow_past_date' => false,
+			'save_twelve_hour_format' => false,
```
westonruter: Let's rename this to just `twelve_hour_format`. The underlying setting should not vary between 12 and 24 hour formats. Only the _display_ would vary.

westonruter: Also, instead of `false` we could use this instead: `false === stripos( get_option( 'time_format' ), 'a' )`.

The default `time_format` is `'g:i a'` so if someone wants to use 24-hour time on their site and change this to `'h:i'`, then the control would automatically use 24-hour time as well.

src/wp-includes/customize/class-wp-customize-date-time-control.php
westonruter: Cool. I've suggested that here: https://github.com/xwp/wordpress-develop/pull/258#discussion_r140137752

src/wp-admin/js/customize-controls.js
```diff
@@ -5025,27 +5553,45 @@
...
 					closeBtn.find( '.screen-reader-text' ).text( api.l10n.cancel );
+					publishSettingsBtn.prop( 'disabled', false );
 				}
 
 				/*
 				 * Save (publish) button should be enabled if saving is not currently happening,
 				 * and if the theme is not active or the changeset exists but is not published.
 				 */
-				canSave = ! saving() && ( ! activated() || ! saved() || ( '' !== changesetStatus() && 'publish' !== changesetStatus() ) );
+				canSave = ! saving() && ( ! activated() || ! saved() || ( changesetStatus() !== selectedChangesetStatus() && '' !== changesetStatus() ) );
```
westonruter: I noticed that if I change the scheduled date it doesn't cause the “Schedule” button to re-enable.

This needs to be updated to take into account when `'future' === selectedChangesetStatus()` and a `selectedChangesetDate()` value is not the same as the most recent `publishDate`. Maybe there needs to be a `changesetDate()` state which reflects the initial and most recently-saved date on the server, similar to how there is a `changesetStatus()` which does the same for the status. Whenever a server update is made, then in addition to how the `status` is passed back we could also pass back the date, and that could get put into the `changesetDate()`. And the condition here could include `changesetDate() !== selectedChangesetStatus()`. The `api.state( 'selectedChangesetDate' )` could then be the `Value` that is supplied to the `changeset_scheduled_date` control.

src/wp-admin/js/customize-controls.js
```diff
@@ -3677,6 +3692,451 @@
...
+					min = parseInt( element.element.attr( 'min' ), 10 );
+					maxLength = parseInt( element.element.attr( 'maxlength' ), 10 );
+					value = parseInt( element(), 10 );
+					control.invalidDate = value > max || value < min || String( value ).length > maxLength || ! _.isEmpty( value );
+					control.toggleErrorNotification( component );
+				}
+			} );
+
+			if ( ! control.params.allowPastDate && ! control.invalidDate ) {
+				control.toggleFutureDateNotification( ! control.isFutureDate() );
```
westonruter: We might want to tone this down or even potentially remove it. Now with https://github.com/xwp/wordpress-develop/pull/258/commits/4f0d53ee1929be560d7bbb96de04cdb43ff0935d when you try to Schedule a post with a paste date it will show the same error notification in the date control as is happening here with `toggleFutureDateNotification`:

image

Getting immediate feedback when I happen to enter an invalid digit can be alarming.

src/wp-admin/js/customize-controls.js
```diff
@@ -3677,6 +3692,451 @@
...
+			}
+		},
+
+		/**
+		 * Toggle error notification for date control.
+		 *
+		 * @param {boolean} notify Add or remove the notification.
+		 * @return {void}
+		 */
+		toggleFutureDateNotification: function toggleErrorNotification( notify ) {
```
westonruter: This needs to be changed from `toggleErrorNotification` to `toggleFutureDateNotification`.

src/wp-admin/js/customize-controls.js
```diff
@@ -3677,6 +3692,451 @@
...
+				var element, max, min, maxLength, value;
+
+				if ( ! control.invalidDate ) {
+					element = control.inputElements[ component ];
+					max = parseInt( element.element.attr( 'max' ), 10 );
+					min = parseInt( element.element.attr( 'min' ), 10 );
+					maxLength = parseInt( element.element.attr( 'maxlength' ), 10 );
+					value = parseInt( element(), 10 );
+					control.invalidDate = value > max || value < min || String( value ).length > maxLength || ! _.isEmpty( value );
+					control.toggleErrorNotification( component );
```
westonruter: Instead of adding a notification, what about instead doing `element.element[0].setCustomValidity( someErrorMessage )` and then in CSS add a rule like:

```css
.date-input:invalid {
    border-color: red;
}
```

Showing something like this:

image

sayedtaqui: Absolutely, I was actually already working on it after this commit, the notification for it is looking a little irritating. 
Another thing I wanted to ask you is that why is the notification not sliding down ?, it just pops open immediately, did we remove the animation or there is something that needs to be done to achieve this?

src/wp-admin/js/customize-controls.js
sayedtaqui: I have seen examples of customizer with a lot of entities where the saving takes some time. In my opinion client side validation is better than doing server side. But to avoid immediate validation I have debounced the check with a delay of 1 second.

```
( _.debounce( function debounceNotification() {
	control.toggleFutureDateNotification( ! control.isFutureDate() );
}, delay ) )();
```
If that still doesn't look good, I will remove that.

sayedtaqui: Or may be show notification when the schedule button is clicked and stop the request if date is not future?

src/wp-admin/js/customize-controls.js
westonruter: Regarding the animation, yes, that was removed in https://github.com/xwp/wordpress-develop/commit/594a41666b5956e23bffd68eef321b613dd65433#diff-58dfe22165ffd9650354f7bdd51002ec

There is an enhancement todo here: 

https://github.com/xwp/wordpress-develop/blob/8436a4f11728bf1c58721ca2a953342f90df0e7c/src/wp-admin/js/customize-controls.js#L184

But even with animation, it would still cause the fields to slide out from under the user. This would be annoying if a user is clicking the `number` input's up/down arrows.

One reason for removing the animation is for the global notification area. The area that global notifications are put sticks them above the panel and the panel then gets positioned `top` at the notifications' `height`. If the notifications have an animated height, then the panel would have to get an animated `top` as well. Ideally this would be done with CSS transitions instead of using jQuery like we were before.

There are also some complications with sticky headers, see https://core.trac.wordpress.org/ticket/41850

Lastly, with how `wp.customize.Notifications` were implemented, it will re-render the entire container now instead of individually add/remove the notifications. If animation were to be re-added, then that would need to change and add complications to the rendering logic.

src/wp-admin/js/customize-controls.js
westonruter: > I have seen examples of customizer with a lot of entities where the saving takes some time.

But wouldn't the save request short-circuit early because of the invalid date? The date validation logic happens before iterating over the settings (entities) in the changeset itself:

https://github.com/xwp/wordpress-develop/blob/52bdeee37d15d8c7f7515013fbcbb6d8bb35d9a6/src/wp-includes/class-wp-customize-manager.php#L2266-L2292

> Or may be show notification when the schedule button is clicked and stop the request if date is not future?

Possibility. If this were a control that had a setting, a way to do that would be:

```js
api.state( 'saving', function( isSaving ) {
    if ( isSaving ) {
        control.toggleFutureDateNotification( ! control.isFutureDate() );
    }
} )
```

This would seem better than listening for clicking on the save button. As long as the notification is added with an `error` type then it will cause the Customizer's `save` to short-circuit here:

https://github.com/xwp/wordpress-develop/blob/52bdeee37d15d8c7f7515013fbcbb6d8bb35d9a6/src/wp-admin/js/customize-controls.js#L4700-L4732

❗️ _However_ this setting doesn't have an associated setting so it won't work. One possibility would be to expand the logic of:

```js
invalidControls = api.findControlsForSettings( invalidSettings );
```

To also loop over all of `api.control` in addition to all the settings to find any (setting-less) controls that have error notifications.

src/wp-admin/js/customize-controls.js
sayedtaqui: And then we wouldn't have to call `api.previewer.save( { date: 'futuredate' } );` explicitly .  The save button `saveBtn.click..` would work for all.

sayedtaqui:
@westonruter I have resolved conflict with master and I think there is no issues, but can you also check and verify if everything is in place? https://github.com/xwp/wordpress-develop/pull/258/commits/84d29856c28f5adc0d9a272804f925a27cd90a0a

There were conflicts in three files namely
`src/wp-includes/class-wp-customize-manager.php`
`src/wp-admin/js/customize-controls.js`
`tests/phpunit/tests/customize/manager.php`

sayedtaqui:
@westonruter 
> So having a z-index: 99999 in the stylesheet is how the allowMultiple section is forced to be displayed over other sections and panels, right? ...

I think that wouldn't be enough, if we want to bring the section on top like for customize-posts plugin, it would require some more css because we dont want to fiddle with the default behaviour of the expand/collapse functionality of the section

```
.section-single-open .customize-pane-child,
.section-single-open .customize-info,
.section-single-open .customize-pane-parent {
	display: none !important;
}

.section-single-open #customize-theme-controls .customize-pane-child.section-single-expanded {
	display: block !important;
	background: #f1f1f1;
	min-height:100%;
	z-index: 99999;
	transform: translate( 0, 0 );
	left: 0;
	top: 0;
	visibility: visible;
}
```

and then in JS

```
var toggleSection = function( sectionId, expand ) {
	api.section( sectionId, function( section ) {
		var fullOverlay = $( '.wp-full-overlay' );
		section.container.toggleClass( 'section-single-expanded', expand );
		fullOverlay.toggleClass( 'section-single-open', expand );
	} );
};
```

and then would also have to put something on top of the back button to execute `toggleSection( 'title_tagline', false )`, because the back button  wouldn't work as it is still closed.  Publish settings section was different case, because we wanted to change its default behaviour.

melchoyce:
Some notes:
- I like this in the expanded panel a lot. Feels way "faster" and more natural this way, IMO.
- It's weird to open "Schedule" and already have an error — can we hold off on showing any error until someone edits the date/time?
- I almost wonder if we should duplicate the "schedule" button inside the date/time box to make it more immediately actionable — thoughts?
- This is super minor, but I don't think we need the spaces in `( "go live" )`
- Also super minor, can we increase the padding on `#customize-control-changeset_scheduled_date` to `padding: 12px 12px 18px;` so it feels more even on top + bottom?

sayedtaqui:
@melchoyce Thanks for the feedback, will work on css and notification.

> I almost wonder if we should duplicate the "schedule" button inside the date/time box to make it more immediately actionable — thoughts?

I am not sure about this one, would let @westonruter comment.

> One more thing I noticed — when I published my changes and then toggled to Save Draft and Schedule, I expected the button to swap labels, even though it's disabled. Any way we can make the label change immediately, even if you don't have changes?

Actually the section should collapse if you publish the changes, and if the status is publish the gear icon should stay disabled until you make any changes because publish settings section is of no use if there are no changes in customizer.
I have yet to connect all these wires, meaning when to disable/enable button when to show/remove notifications, when to change button text.  I was planning to have the components ready before doing it.

westonruter:
> I almost wonder if we should duplicate the "schedule" button inside the date/time box to make it more immediately actionable — thoughts?

@melchoyce If you think that redundancy is helpful. But otherwise, I'd think the changing of the text in the button and any associated switch from disabled to enabled should cause it to pop out as being the next thing to click. Maybe additional highlighting should be added to that button in this case and in other scenarios, like if you click to exit the Customizer but then click Cancel, so the user can have their eyes drawn to the button for what they failed to press. This seems better maybe than adding a secondary Schedule button which is another place to do the same thing.

melchoyce:
Maybe we could make the button shake if they have changes to schedule, and they close the publish settings panel? @karmatosed any thoughts here?

karmatosed:
Overall amazing work everyone. I just wanted to drop in and say after giving it a run through, I am very impressed with what you've got in.

+1 to animation. Some examples I've played with in past:

https://codepen.io/karmatosed/pen/meeNoM
https://codepen.io/karmatosed/pen/vNNZYb

I am wondering about the language 'go live'. For example:

"Schedule your customization changes to publish ( "go live" ) at a future date."

I know users may not understand publish, do they understand go live?

I also feel the text + message here is a little long:
2017-09-25 at 20 50

That's a lot of words, thinking of translation that is going to get quite intense.

src/wp-admin/js/customize-controls.js
```diff
@@ -1137,8 +1139,9 @@
 		 */
 		embed: function () {
 			var inject,
-				section = this,
-				container = $( '#customize-theme-controls' );
+				section = this;
+
+			section.cointainerParent = api.ensure( section.containerParent );
```
westonruter: Misspelled: `cointainerParent/containerParent `

src/wp-admin/js/customize-controls.js
```diff
@@ -5406,6 +5433,9 @@
 		api.section( 'publish_settings', function( section ) {
 			var backgroundEls, animationDuration = 500, updateButtonsState;
 
+			section.containerPaneParent = '.customize-outer-pane-parent';
+			section.containerParent = '#customize-outer-theme-controls';
```
westonruter: Isn't this redundant with what is in `OuterSection#initialize`?

sayedtaqui: Yes it is redundant, have to remove it.

sayedtaqui: In fact there are more things which needs to be removed from `publish_settings` section as they would be handled within outerSection

sayedtaqui:
The outerSection can now support any number of sections, by just adding `type => 'outer'` when registering section in php. Probably in future the widgets and menu outer sections can also make use of `api.outerSection` which will cut down some code, or may be new sections which needs to be outside the sidebar. You can test it by adding some sections and controls to see how it works.

```php
$this->add_section( 'section_test_1', array(
	'title' => __( 'Section Test 1' ),
	'priority' => 0,
	'capability' => 'customize',
	'type' => 'outer',
) );

$this->add_control( 'section_test_1_control_1', array(
	'section' => 'section_test_1',
	'settings' => array(),
	'type' => 'radio',
	'choices' => array(
		'rat' => 'rat',
		'horse' => 'horse',
		'goat' => 'goat',
	),
	'label' => __( 'Control 1' ),
	'capability' => 'customize',
) );

$this->add_control( 'section_test_1_control_2', array(
	'section' => 'section_test_1',
	'settings' => array(),
	'type' => 'text',
	'label' => __( 'Control 2' ),
	'capability' => 'customize',
) );

$this->add_section( 'section_test_2', array(
	'title' => __( 'Section Test 2' ),
	'priority' => 0,
	'capability' => 'customize',
	'type' => 'outer',
) );
```

This is meant to show only one section at a time and can be triggered using

```js
wp.customize.section( 'section_test_1' ).expand();
wp.customize.section( 'section_test_1' ).collapse();
```
This is just to let you now that it in place and It would probably need more testing and finishing which we can do later.

src/wp-includes/customize/class-wp-customize-date-time-control.php
```diff
@@ -0,0 +1,250 @@
...
+		<span class="description customize-control-description">{{ data.description }}</span>
+		<div class="date-time-fields">
+			<div class="day-row">
+				<span class="title-day"><?php esc_html_e( 'Day' ); ?></span>
+				<div class="day-fields clear">
+					<label class="month-field">
+						<span class="screen-reader-text"><?php esc_html_e( 'Month' ); ?></span>
+							<select class="date-input month" data-component="month">
+								<# _.each( data.month_choices, function( choice ) {
+										if ( _.isObject( choice ) && ! _.isUndefined( choice.text ) && ! _.isUndefined( choice.value ) ) {
```
westonruter: Looks like too much indentation here.

src/wp-includes/customize/class-wp-customize-date-time-control.php
```diff
@@ -0,0 +1,250 @@
...
+						<input type="number" size="4" maxlength="{{ data.maxYearLength }}" autocomplete="off" class="date-input year" data-component="year" min="{{ data.minYear }}" value="{{ data.year }}" max="{{ data.maxYear }}" />
+					</label>
+				</div>
+			</div>
+			<div class="time-row clear">
+				<span class="title-time"><?php esc_html_e( 'Time' ); ?></span>
+				<div class="time-fields clear">
+					<label class="hour-field">
+						<span class="screen-reader-text"><?php esc_html_e( 'Hour' ); ?></span>
+						<# data.maxHour = data.twelveHourFormat ? 12 : 24; #>
```
westonruter: Instead of amending `maxHour` onto `data`, let's just `var maxHour`. Mutating the `data` object could cause side effects.

src/wp-includes/customize/class-wp-customize-date-time-control.php
```diff
@@ -0,0 +1,250 @@
...
+					<# if ( data.twelveHourFormat ) { #>
+					<label class="am-pm-field">
+						<span class="screen-reader-text"><?php esc_html_e( 'AM / PM' ); ?></span>
+						<select class="date-input" data-component="ampm">
+							<option value="am"><?php esc_html_e( 'AM' ); ?></option>
+							<option value="pm"><?php esc_html_e( 'PM' ); ?></option>
+						</select>
+					</label>
+					<# } #>
+					<span class="date-timezone" aria-label="<?php esc_attr_e( 'Timezone' ); ?>" title="<?php echo esc_attr( $timezone_info['description'] ); ?>"><?php echo esc_html( $timezone_info['abbr'] ); ?></span>
```
westonruter: Should this be an actual `abbr` element?

src/wp-includes/customize/class-wp-customize-date-time-control.php
```diff
@@ -0,0 +1,250 @@
...
+	 *
+	 * @return array abbr and description.
+	 */
+	public function get_timezone_info() {
+		$tz_string = get_option( 'timezone_string' );
+		$timezone_info = array();
+
+		if ( $tz_string ) {
+			$tz = new DateTimezone( $tz_string );
+			$now = new DateTime( 'now', $tz );
```
westonruter: We should probably wrap these in `try`/`catch` in case the `timezone_string` is not valid.

src/wp-includes/customize/class-wp-customize-preview-link-control.php
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+/**
+ * Customize Preview Link Control class.
+ *
+ * @since 4.9.0
+ *
+ * @see WP_Customize_Control
+ */
+class WP_Customize_Preview_Link_Control extends WP_Customize_Control {
```
westonruter: I wonder if we should really create a new PHP class or this since it's really a single-use kind of thing. It may be better to just output the template into the page and then register the control type in JS “manually”.

src/wp-includes/customize/class-wp-customize-date-time-control.php
```diff
@@ -0,0 +1,250 @@
...
+							</select>
+					</label>
+					<label class="day-field">
+						<span class="screen-reader-text"><?php esc_html_e( 'Day' ); ?></span>
+						<input type="number" size="2" maxlength="2" autocomplete="off" class="date-input day" data-component="day" min="1" max="31" value="{{ data.day }}" />
+					</label>
+					<span class="time-special-char date-time-separator">,</span>
+					<label class="year-field">
+						<span class="screen-reader-text"><?php esc_html_e( 'Year' ); ?></span>
+						<#  data.maxYearLength = String( data.maxYear ).length; #>
```
westonruter: An extra space appears at the beginning here.

westonruter: Also, I think we can remove this entirely and hard-code the `maxlength` of 4. It's not like we're anticipating this code to be running still in 7,983 years 😂

src/wp-includes/customize/class-wp-customize-date-time-control.php
```diff
@@ -0,0 +1,250 @@
...
+										#>
+								<option value="{{ value }}" {{selected}} >
+									{{ text }}
+								</option>
+								<# } ); #>
+							</select>
+					</label>
+					<label class="day-field">
+						<span class="screen-reader-text"><?php esc_html_e( 'Day' ); ?></span>
+						<input type="number" size="2" maxlength="2" autocomplete="off" class="date-input day" data-component="day" min="1" max="31" value="{{ data.day }}" />
```
westonruter: I see hard-coding of the values in this template. Shouldn't this be handled by `wp.customize.Element` automatically?

src/wp-admin/js/customize-controls.js
```diff
@@ -1840,6 +1850,133 @@
...
+		attachEvents: function() {
+			var section = this, body = $( 'body' );
+
+			section.expanded.bind( function( isExpanded ) {
+				body.toggleClass( 'outer-section-open', isExpanded );
+				section.container.toggleClass( 'open', isExpanded );
+				section.container.removeClass( 'busy' );
+				api.section.each( function( _section ) {
+				    if ( 'outer' === _section.params.type && _section.id !== section.id ) {
+						_section.container.removeClass( 'open' );
```
westonruter: Extra tab appears here.

src/wp-admin/js/customize-controls.js
```diff
@@ -1840,6 +1850,133 @@
...
+
+		/**
+		 * Overrides api.Section.prototype.attachEvents to add behaviors for outer accordion section.
+		 *
+		 * @since 4.9.0
+		 */
+		attachEvents: function() {
+			var section = this, body = $( 'body' );
+
+			section.expanded.bind( function( isExpanded ) {
```
westonruter: Should this be moved into `onChangeExpanded`?

src/wp-admin/js/customize-controls.js
```diff
@@ -3960,6 +4097,512 @@
...
+					el = element.element.get( 0 );
+					max = parseInt( element.element.attr( 'max' ), 10 );
+					min = parseInt( element.element.attr( 'min' ), 10 );
+					maxLength = parseInt( element.element.attr( 'maxlength' ), 10 );
+					value = element();
+					control.invalidDate = value > max || value < min || String( value ).length > maxLength;
+					errorMessage = control.invalidDate ? api.l10n.invalid + ' ' + component : '';
+
+					el.setCustomValidity( errorMessage );
+					_.result( el, 'reportValidity' );
```
westonruter: Interesting. I've never used this `_.result` function before.

src/wp-admin/js/customize-controls.js
```diff
@@ -3960,6 +4097,512 @@
...
+		 */
+		updateDaysForMonth: function updateDaysForMonth() {
+			var control = this, daysInMonth, year, month, day;
+
+			month = control.inputElements.month();
+			year = control.inputElements.year();
+			day = control.inputElements.day();
+
+			if ( month && year ) {
+				daysInMonth = new Date( year, month, 0 ).getDate();
```
westonruter: Nice technique.

src/wp-admin/js/customize-controls.js
```diff
@@ -3960,6 +4097,512 @@
 		}
 	});
 
+	/**
+	 * Class wp.customize.DateTimeControl.
+	 *
```
westonruter: Let's add `@since` tags to the class and all methods.

src/wp-includes/customize/class-wp-customize-date-time-control.php
westonruter: This should be handled automatically by `populateDateInputs`.

src/wp-admin/js/customize-controls.js
```diff
@@ -3960,6 +4097,512 @@
...
+	});
+
+	/**
+	 * Class wp.customize.PreviewLinkControl.
+	 *
+	 * @constructor
+	 * @augments wp.customize.Control
+	 * @augments wp.customize.Class
+	 */
+	api.PreviewLinkControl = api.Control.extend({
```
westonruter: As with perhaps opting to not add a new PHP class for an ad hoc control, maybe too this can be `var PreviewLinkControl` and so not exported. The control could be added dynamically in the `publish_settings` section.

westonruter: Humm, but then I suppose we wouldn't be able to use PHP to manage the registration of the control. But would we even need that?

If a plugin wanted to disable the ability to schedule changesets, they could remove the `future` choice from status dropdown control, and that could be a signal that this `PreviewLinkControl` could be skipped entirely from being added.

src/wp-admin/js/customize-controls.js
```diff
@@ -6221,6 +7037,75 @@
...
+										api.notifications.add( code, notification );
+										api.settings.changeset.uuid = resp.next_changeset_uuid
+									}
+								}
+							} );
+						} );
+					};
+
+					publishWhenTime( api.state( 'changesetDate' ).get() );
+					api.state( 'changesetDate' ).bind( publishWhenTime );
```
westonruter: @sayedtaqui won't this call `publishWhenTime` every time the changeset changes and then result in another `timeArrived` promise being added? If you change the future date a bunch of times, this seems like it could cause X number or promises and calls to `requestChangesetUpdate`.

src/wp-admin/js/customize-controls.js
```diff
@@ -6221,6 +7037,75 @@
...
+					publishWhenTime = function( datetime ) {
+						if ( 'future' !== api.state( 'changesetStatus' ).get() || ! datetime ) {
+						    return;
+						}
+
+						dateControl.timeArrived( datetime, timeInterval ).done( function() {
+							var request, notification, code = 'scheduled_changeset_published';
+
+							request = api.requestChangesetUpdate( {}, {
+								title: api.settings.changeset.uuid
```
westonruter: I'm not sure we should include the title here.

src/wp-admin/js/customize-controls.js
```diff
@@ -6221,6 +7037,75 @@
...
+							} );
+
+							request.done( function( resp ) {
```
westonruter: I would expect that actually `request` would fail if the changeset was published. Wouldn't it return `changeset_already_published` code in an error response?

src/wp-admin/js/customize-controls.js
```diff
@@ -6221,6 +7037,75 @@
...
+								if ( 'trash' === resp.changeset_status || 'changeset_already_published' === resp.code ) {
+									api.state( 'changesetStatus' ).set( '' );
+									if ( resp.next_changeset_uuid ) {
+										notification = new api.Notification( code, {
+											message: api.l10n.changesetPublished,
+											type: 'info',
+											dismissible: true
+										} );
+										api.notifications.add( code, notification );
+										api.settings.changeset.uuid = resp.next_changeset_uuid
```
westonruter: We'll also need to send a `saved` message to the preview to include the `next_changeset_uuid`.

src/wp-admin/js/customize-controls.js
sayedtaqui: `changesetDate ` would only be triggered when the date is saved from the server, we have a different state for selected date `selectedChangesetDate`. If you meant `changesetDate`,  I would wonder how many times would someone reschedule there changeset.

src/wp-admin/js/customize-controls.js
sayedtaqui: So if do not send title and if there are no dirty changes, it will short circuit and not send an ajax request.

src/wp-admin/js/customize-controls.js
sayedtaqui: Yes, it will, don't how we will update the uuid then.

src/wp-includes/customize/class-wp-customize-date-time-control.php
sayedtaqui: Meaning `max`?, actually that is being updated in js in `api.DateTimeControl.updateDaysForMonth`, other values don't need update.

src/wp-includes/customize/class-wp-customize-date-time-control.php
sayedtaqui: Well, that's right but since this date control is not specific to schedule date control only and can be used by anyone and I have given the option of `max_year` and `min_year`,  so I cannot hard code the length unless I remove this option of `max_year`, which I think would not be useful for others.

src/wp-includes/customize/class-wp-customize-date-time-control.php
sayedtaqui: I think yes.

src/wp-admin/js/customize-controls.js
sayedtaqui: Its same as doing 
```js
if ( 'undefined' !== typeof el.reportValidity ) {
	el.reportValidity();
}
```

src/wp-admin/js/customize-controls.js
westonruter: OK, yes, I missed `selectedChangesetDate` vs `changesetDate` here. Even so, however, if I first save a scheduled changeset to go live in 1 hour, but then I change my mind to schedule it to go live in 2 hours, as I understand the first 1-hour `publishWhenTime` will still be pending and will resolve when that hour is up. True that if the date is changed then `requestChangesetUpdate` would essentially do nothing, but also since it lacks `autosave` then it will write any pending changes into the full changeset and not a revision.

I think this needs some more thinking.

src/wp-admin/js/customize-controls.js
westonruter: I see. OK so then it's somewhat of a hack. I'm trying to think of a better way.

westonruter: We could instead introduce a `force` param here.

westonruter: Or we could just use `wp.customize.previewer.save()` instead. That would be better I think, as then all of the events related to saving would trigger, along with grabbing the new changeset UUID.

src/wp-admin/js/customize-controls.js
westonruter: I see that there won't be a possibility for concurrency due to the code you have here:

```js
clearInterval( control.interval ); // To ensure multiple intervals are not running in any case.
```

src/wp-admin/js/customize-controls.js
westonruter: I've implemented this with a `changeset-uuid` message.

src/wp-admin/js/customize-controls.js
westonruter: I ended up refactoring this to use a single interval that is managed by the Publish Settings section (since there are dependencies with other states), instead of using promises inside of the control.

src/wp-includes/customize/class-wp-customize-date-time-control.php
westonruter: No, I mean `value="{{ data.day }}"`. I believe you should be able to omit all of `value` attributes and even `selected` because all of this will be handled by `wp.customize.Element`.

src/wp-admin/js/customize-controls.js
sayedtaqui: > If a plugin wanted to disable the ability to schedule changesets, they could remove the future choice from status dropdown control, and that could be a signal that this PreviewLinkControl could be skipped entirely from being added.

If the future status is removed, why would `PreviewLinkControl` be skipped, it can show the preview link of draft changeset isn't?

src/wp-admin/js/customize-controls.js
```diff
@@ -7052,8 +7053,8 @@
 
 					if ( shouldPoll && ! pollInterval ) {
 						pollInterval = setInterval( function() {
-							var remainingTime = api.utils.getRemainingTime( api.state( 'changesetDate' ).get() );
-							if ( remainingTime <= 0 ) {
+							api.state( 'remainingTimeToPublish' ).set( api.utils.getRemainingTime( api.state( 'changesetDate' ).get() ) );
```
sayedtaqui: @westonruter I added a new state `remainingTimeToPublish` because it would be needed to run the countdown in snapshot plugin.

src/wp-admin/js/customize-controls.js
sayedtaqui: @westonruter I think this is resolved now, correct?

src/wp-admin/js/customize-controls.js
```diff
@@ -7009,9 +7009,9 @@
 		/**
 		 * Publish settings section and controls.
 		 */
-		api.control( 'changeset_status', function( statusControl ) {
-			statusControl.deferred.embedded.done( function() {
-				var radioNodes, statusElement;
+		api.control( 'changeset_status', 'changeset_scheduled_date', function( statusControl, dateControl ) {
+			$.when( statusControl.deferred.embedded, dateControl.deferred.embedded ).done( function() {
```
sayedtaqui: Excellent 👍

src/wp-admin/js/customize-controls.js
```diff
@@ -1148,19 +1200,19 @@
...
-								container.append( section.contentContainer );
+								section.containerParent.append( section.contentContainer );
 							}
 							section.deferred.embedded.resolve();
 						});
 					} );
 				} else {
 					// There is no panel, so embed the section in the root of the customizer
-					parentContainer = $( '.customize-pane-parent' ); // @todo This should be defined elsewhere, and to be configurable
+					parentContainer = api.ensure( section.containerPaneParent );
```
westonruter: Great work here!

westonruter:
Landed in [r41626](https://core.trac.wordpress.org/changeset/41626) 🎉

PR #259, opened 2017-09-18
Trac 39778: Add wp_is_uuid() function

westonruter:
https://core.trac.wordpress.org/ticket/39778


westonruter:
Committed to trunk in https://core.trac.wordpress.org/changeset/41388

PR #260, opened 2017-09-19
Trac #38794: Customize: Allow panels and sections to have notifications just like controls can

westonruter:
https://core.trac.wordpress.org/ticket/38794


westonruter:
Committed in r41390.

PR #267, opened 2017-10-01
Trac 39896: Trashing/reverting changesets and introduce OverlayNotification

westonruter:
Changes have been deployed to http://trac-39896-xwp-testbed.pantheonsite.io/wp-admin/customize.php

This PR does various cleanup of code related to drafting/scheduling and the new experience for themes.

* Add a new trash button to Publish Settings section that invokes a new `wp.customize.previewer.trash()` to trash the current changeset. Add logic to `WP_Customize_Manager` to handle deleting changeset drafts. Add `trashing` to `wp.customize.state` which is then used to update the UI.
* Improve logic for managing the visibility and disabled states for publish buttons.
* Prevent attempting `requestChangesetUpdate` while processing and bump processing while doing `save`.
* Only update `changeset_date` state if sent in save response.
* Merge `ThemesSection#loadThemePreview()` into `ThemesPanel#loadThemePreview()`.
* Remove unused `autosaved` state.
* Start autosaving and prompting at beforeunload after a change first happens. This is key for theme previews since even if a user did not make any changes, there were still dirty settings which would get stored in an auto-draft unexpectedly.
* Allow `Notification` to accept additional `classes` to be added to `container`.
* Introduce `OverlayNotification` and use for theme installing, previewing, and trashing. Such overlay notifications take over the entire window.

- [x] Add unit tests for trashing.
- [ ] Add keyboard-accessibility to publish settings (needs tabindex work).

See:

* https://core.trac.wordpress.org/ticket/39896
* https://core.trac.wordpress.org/ticket/37661
* https://core.trac.wordpress.org/ticket/21666

### DIFF
--- a/src/wp-admin/js/customize-controls.js
+++ b/src/wp-admin/js/customize-controls.js
@@ -2925,7 +2925,7 @@
 		 * @returns {jQuery.promise} Promise.
 		 */
 		loadThemePreview: function( themeId ) {
-			var deferred = $.Deferred(), onceProcessingComplete, overlay, urlParser, queryParams;
+			var deferred = $.Deferred(), onceProcessingComplete, urlParser, queryParams;
 
 			urlParser = document.createElement( 'a' );
 			urlParser.href = location.href;
--- a/src/wp-admin/js/customize-controls.js
+++ b/src/wp-admin/js/customize-controls.js
@@ -6238,6 +6238,7 @@
 		'saved',
 		'autosaved',
 		'saving',
+		'trashing',
 		'activated',
 		'processing',
 		'paneVisible',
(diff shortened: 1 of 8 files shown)


src/wp-admin/js/customize-controls.js
```diff
@@ -2899,8 +2906,13 @@
 			// Also preview the theme as the event is triggered on Install & Preview.
 			if ( $( event.target ).hasClass( 'preview' ) ) {
 				preview = true;
-				$( '.wp-full-overlay' ).addClass( 'customize-loading' );
-				wp.a11y.speak( $( '#customize-themes-loading-container .customize-loading-text-installing-theme' ).text() );
+
+				api.notifications.add( 'theme_installing', new api.OverlayNotification( 'theme_installing', {
+					message: api.l10n.themeDownloading,
+					dismissible: true,
```
celloexpressions: It looks like this doesn't actually work for theme installing, as the reload still happens after the install is complete. I'm fine with this not being dismissible, unless someone has time to fix that.

westonruter: Good catch. I actually meant to remove this!

westonruter:
Committed in r41667.

PR #269, opened 2017-10-02
Trac-42024 : Add Customization Locking

sayedtaqui:
https://core.trac.wordpress.org/ticket/42024

Todos:

- [x]  Improve `check_changeset_lock` to return false, if the changes were saved before 150s, to handle the case when user has just closed the window.
- [x] Refresh the saved time with heartbeat, so that `check_changeset_lock` does not return false if user is just sitting on customizer ( after doing the above task ).
- [x] Remove changeset lock if user has navigated away from customizer.
- [x] Remove the global notification when user tried to save something however it was taken over by another user meanwhile and hence not saved in favour of the lock notice.
- [x] If user has not refreshed the page after seeing notice, do not change it with take over notice with heartbeat.
- [x] Check how the default locking will work when changeset post type `show_ui` is `true`
- [x] Constrain tabs to lock window.
- [x] Change user_name keys to display_name
- [x] Confirm Firefox compat ?
- [ ] Write Tests.
- [x] Use custom-styled `OverlayNotification` instead of custom element. This would add tab constraining for free.
- [x] Handle case for user was taken over but have unsaved changes. There is currently an AYS dialog when clicking on Preview but there is nothing they can do about it. What should happen with the user's autosave revision? Should it be deleted? If it is prompted when they return to the Customizer, should the updated changeset get merged on top of the autosave revision? Clicking Preview should cause an autosave to happen before the leave and prevent the AYS dialog. 
- [x] Or if the changeset was published by the other user, any autosaves remaining on that published changeset could get converted into autosave revisions.
- [ ] Then when they come back, and they click to restore the autosave, we can merge it on top of the changes the other user made. See https://core.trac.wordpress.org/ticket/42191


src/wp-includes/class-wp-customize-manager.php
```diff
@@ -2919,6 +2951,248 @@ public function grant_edit_post_capability_for_changeset( $caps, $cap, $user_id,
...
+						<div class="customize-changeset-locked-avatar"><?php echo $user_avatar; ?></div>
+						<p class="currently-editing wp-tab-first" tabindex="0">
+							<?php printf( '<span class="customize-notice-user-name">%s</span> <span class="customize-take-over-message">%s</span>', $user_name, esc_html__( 'is already customizing this site. Do you want to take over?' ) ); ?>
+						</p>
+						<p>
+							<?php if ( $this->get_preview_url() !== $this->get_return_url() ) { ?>
+								<a class="button customize-notice-go-back-button" href="<?php echo esc_url( $this->get_return_url() ); ?>"><?php esc_html_e( 'Go back' ); ?></a>
+							<?php } ?>
+							<a class="button customize-notice-preview-button" href="<?php echo esc_url( $preview_url ); ?>"><?php esc_html_e( 'Preview' ); ?></a>
+							<a class="button button-primary wp-tab-last customize-notice-take-over-button" href="<?php echo $take_over; ?>"><?php esc_html_e( 'Take over' ); ?></a>
```
westonruter: I should this not be a regular `button` which then causes an Ajax request to lock the changeset? Then the modal can just be closed without having to reload the Customizer entirely.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -2919,6 +2951,248 @@ public function grant_edit_post_capability_for_changeset( $caps, $cap, $user_id,
...
+		wp_redirect( $redirect_url );
+		exit();
+	}
+
+	/**
+	 * Render template for changeset locked notice.
+	 *
+	 * @since 4.9.0
+	 */
+	public function render_changeset_locked_notice_template() {
```
westonruter: I think we can just as well put the template into `render_control_templates` like we have for other general purpose templates.

src/wp-includes/class-wp-customize-manager.php
westonruter: In particular, the request should result in calling `take_over_changeset_on_request`, and then do `api.state( 
'changesetLocked' ).set( false )`

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -2919,6 +2951,248 @@ public function grant_edit_post_capability_for_changeset( $caps, $cap, $user_id,
...
+		$hidden_class = 'hidden';
+		$query_args = array_merge( array(
+			'changeset_uuid' => $this->changeset_uuid(),
+			'action' => $this->_changeset_take_over_action,
+			'has_changeset_uuid_param' => ! empty( $_GET['changeset_uuid'] ),
+			'nonce' => wp_create_nonce( $this->_changeset_take_over_action ),
+		), $this->get_autofocus() );
+
+		if ( isset( $_GET['return'] ) ) {
+			$query_args['return'] = $_GET['return'];
```
westonruter: This should be a call to `$wp_customize->get_return_url()`, right? No need to access `$_GET['return']` directly.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -2919,6 +2951,248 @@ public function grant_edit_post_capability_for_changeset( $caps, $cap, $user_id,
...
+		}
+
+		if ( isset( $_GET['url'] ) ) {
+			$query_args['url'] = $_GET['url'];
```
westonruter: This should be a call to `$wp_customize->get_preview_url()`, right? No need to access `$_GET['url']` directly.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -2919,6 +2951,248 @@ public function grant_edit_post_capability_for_changeset( $caps, $cap, $user_id,
...
+
+		return false;
+	}
+
+	/**
+	 * Removes changeset lock when take over request is sent.
+	 *
+	 * @since 4.9.0
+	 */
+	public function take_over_changeset_on_request() {
```
westonruter: I think this should be an Ajax handler instead.

src/wp-includes/script-loader.php
```diff
@@ -564,6 +564,7 @@ function wp_default_scripts( &$scripts ) {
 		'saved'              => __( 'Saved' ),
 		'cancel'             => __( 'Cancel' ),
 		'close'              => __( 'Close' ),
+		'takenOverMessage'   => __( 'has taken over and is currently customizing.' ),
```
westonruter: This seems like it would be hard to translate. Shouldn't there be a placeholder for the subject?

src/wp-includes/class-wp-customize-manager.php
sayedtaqui: Yes, ajax would be easier, but I was trying to remove any wait time ajax call might have, and also to follow how its done for post locking. Shall I use ajax?

src/wp-admin/js/customize-controls.js
```diff
@@ -7660,6 +7674,76 @@
...
+		/**
+		 * Handles lock notice and take over request.
+		 *
+		 * @since 4.9.0
+		 */
+		( function checkAndDisplayLockNotice() {
+			var template, body, request, takeOverButton, errorMessage, focusAndToggleLockNotice;
+
+			body = $( 'body' );
+			template = $( wp.template( 'customize-changeset-locked-notice' )() );
```
westonruter: I want to try using a custom-styled `OverlayNotification` for this.

westonruter:
Committed in r41839.

PR #272, opened 2017-10-03
Trac 21622: Validate or sandbox theme file edits before saving them (as is done for plugins)

westonruter:
See https://core.trac.wordpress.org/ticket/21622

* Refactor themes and plugin editors to submit file changes via Ajax instead of doing full page refreshes. Moves file modifications from `theme-editor.php` and `plugin-editor.php` to common Ajax handler.
* Do sandboxed updates of PHP files for both active themes and plugins, with rollbacks in the case of a loopback request results in a fatal error. This supersedes changes in [#39766](https://core.trac.wordpress.org/ticket/39766) (Plugin does not gracefully fail when editing active plugin causes fatal error).
* Use `get` method for plugin selection dropdown.

Todo:

- [x] Remove duplicated code in Ajax handler and the `theme-editor.php` and `plugin-editor.php` files.
- [x] Reliably scrape PHP errors that get generated in the loopback requests, forcing the display of errors in the request.
- [x] Ensure removed code from `theme-editor.php` and `plugin-editor.php` is all accounted for in Ajax handler.


westonruter:
@melchoyce in this PR I've constrained the height of the file list to be the same as the editor, and in the plugin editor I've stripped the plugin directory name from being included (see #24048):

image

westonruter:
Example of PHP fatal error being caught and rolled back:

image

westonruter:
Committed in r41721.

PR #274, opened 2017-10-04
Clean up longstanding annoyances about the Customize JS API

westonruter:
See https://core.trac.wordpress.org/ticket/30741


westonruter:
Committed in r41726.

PR #275, opened 2017-10-06
Increase size of menu and widget buttons to make them more noticeable

melchoyce:
See https://core.trac.wordpress.org/ticket/41611.

### DIFF
--- a/src/wp-admin/css/customize-controls.css
+++ b/src/wp-admin/css/customize-controls.css
@@ -2337,8 +2337,11 @@ body.cheatin p {
 #customize-theme-controls .add-new-menu-item {
 	cursor: pointer;
 	float: right;
+    font-weight: 600;
 	margin: 0;
 	margin-left: 10px;
+    padding: 5px 15px;
+    height: auto;
 	transition: all 0.2s;
 	-webkit-user-select: none;
 	-moz-user-select: none;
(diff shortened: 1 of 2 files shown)


src/wp-admin/css/customize-controls.css
```diff
@@ -2337,8 +2337,11 @@ body.cheatin p {
 #customize-theme-controls .add-new-menu-item {
 	cursor: pointer;
 	float: right;
+    font-weight: 600;
 	margin: 0;
 	margin-left: 10px;
+    padding: 5px 15px;
+    height: auto;
```
westonruter: Looks like spaces may be used here instead of tabs.

westonruter:
(Ignore the unrelated build errors.)

src/wp-admin/css/customize-controls.css
melchoyce: Weird, looks normal in my editor. I'll copy and paste the tabs directly.

src/wp-admin/css/customize-controls.css
```diff
@@ -2337,8 +2337,11 @@ body.cheatin p {
 #customize-theme-controls .add-new-menu-item {
 	cursor: pointer;
 	float: right;
+    font-weight: 600;
```
westonruter: Spaces here too :)

westonruter:
I'm setting the button not looking quite right on mobile:

image
image

PR #277, opened 2017-10-06
#42126: Customize: Fix confusion related to visibility of Themes panel and drafted/scheduled changesets

westonruter:
https://core.trac.wordpress.org/ticket/42126


westonruter:
Committed in r41788.

PR #278, opened 2017-10-08
#40254: Eliminate use of customize-loader

westonruter:
https://core.trac.wordpress.org/ticket/40254

- [x] When the Customizer reloads due to theme switches or trashing of changesets, make sure the original `return` param gets inserted into the URL that is replaced as otherwise the referer is lost.
- [x] Update `themes.php` to hide initial list of themes until JS initializes to prevent momentary flash of non-filtered themes. This is only necessary when the `search` param is present.
- [x] Add back-compat for customize-loader with the AYS dialog and dismissing autosaves.

### DIFF
--- a/src/wp-admin/css/themes.css
+++ b/src/wp-admin/css/themes.css
@@ -7,6 +7,10 @@
   16.1 - Manage Themes
 ------------------------------------------------------------------------------*/
 
+body.js .theme-browser.search-loading {
+	display: none;
+}
+
 .theme-browser .themes {
 	clear: both;
 }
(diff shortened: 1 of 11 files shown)


westonruter:
@sayedtaqui see https://github.com/xwp/wordpress-develop/pull/278/commits/66977f280e5d23eb3e411917f349b47d201d9f94 specifically for where the customization lock could be lifted if a plugin by chance made use of the customizer-loader, even though we're discontinuing its use in core.

sayedtaqui:
Looks good, just noticed one minor difference in testing which I think is probably intentional? 
If I `wp_enqueue_script( 'customize-loader' )` on dashboard again and I load customizer with `wp.customize.Loader.open( 'http://wp-develop.dev/wp-admin/customize.php' );` from console , 
 the `customize_uuid` param is appended to the customizer url, after its loaded, ( branching is not true ).

westonruter:
Committed in r41797.

PR #281, opened 2017-10-11
#24048: Code Editors: Increase the usability of Code Editor's files list

westonruter:
https://core.trac.wordpress.org/ticket/24048


src/wp-admin/includes/misc.php
```diff
@@ -270,6 +270,180 @@ function update_recently_edited( $file ) {
...
+}
+
+/**
+ * Outputs the formatted file list for the Theme Editor.
+ *
+ * @since 4.9.0
+ * @access private
+ *
+ * @param array  $tree  List of file/folder paths.
+ * @param string $label Name of file or folder to print.
```
westonruter: This is a string here but it has a default value of `false` in the function definition.

src/wp-admin/includes/misc.php
```diff
@@ -270,6 +270,180 @@ function update_recently_edited( $file ) {
...
+}
+
+/**
+ * Outputs the formatted file list for the Plugin Editor.
+ *
+ * @since 4.9.0
+ * @access private
+ *
+ * @param array  $tree  List of file/folder paths.
+ * @param string $label Name of file or folder to print.
```
westonruter: This is a string here but it has a default value of `false` in the function definition.

src/wp-admin/plugin-editor.php
```diff
@@ -144,6 +144,9 @@
 	wp_add_inline_script( 'wp-theme-plugin-editor', sprintf( 'jQuery( function( $ ) { wp.themePluginEditor.init( $( "#template" ), %s ); } )', wp_json_encode( $settings ) ) );
 	wp_add_inline_script( 'wp-theme-plugin-editor', sprintf( 'wp.themePluginEditor.themeOrPlugin = "plugin";' ) );
 
+	wp_enqueue_script( 'tree-links' );
+	wp_add_inline_script( 'tree-links', 'jQuery( function( $ ) { $(\'#templateside [role="group"]\').parent().attr(\'aria-expanded\', false); $(\'#templateside .notice\').parents(\'[aria-expanded]\').attr(\'aria-expanded\', true); } )' );
```
westonruter: I think this logic should be added to `theme-plugin-editor.js`, with either `tree-links` added as a dependency for that script or the code for tree-links to be bundled with `theme-plugin-editor` if we're not going to use it anywhere else.

src/wp-admin/theme-editor.php
```diff
@@ -129,6 +137,9 @@
 	wp_add_inline_script( 'wp-theme-plugin-editor', sprintf( 'jQuery( function( $ ) { wp.themePluginEditor.init( $( "#template" ), %s ); } )', wp_json_encode( $settings ) ) );
 	wp_add_inline_script( 'wp-theme-plugin-editor', 'wp.themePluginEditor.themeOrPlugin = "theme";' );
 
+	wp_enqueue_script( 'tree-links' );
+	wp_add_inline_script( 'tree-links', 'jQuery( function( $ ) { $(\'#templateside [role="group"]\').parent().attr(\'aria-expanded\', false); $(\'#templateside .notice\').parents(\'[aria-expanded]\').attr(\'aria-expanded\', true); } )' );
```
westonruter: I think this logic should be added to `theme-plugin-editor.js`, with either `tree-links` added as a dependency for that script or the code for tree-links to be bundled with `theme-plugin-editor` if we're not going to use it anywhere else.

src/wp-includes/script-loader.php
```diff
@@ -886,6 +886,9 @@ function wp_default_scripts( &$scripts ) {
 		$scripts->add( 'media-gallery', "/wp-admin/js/media-gallery$suffix.js", array('jquery'), false, 1 );
 
 		$scripts->add( 'svg-painter', '/wp-admin/js/svg-painter.js', array( 'jquery' ), false, 1 );
+
+		$scripts->add( 'treeitem-links', "/wp-admin/js/treeitem-links$suffix.js", array( 'jquery' ), false, 1 );
+		$scripts->add( 'tree-links', "/wp-admin/js/tree-links$suffix.js", array( 'treeitem-links' ), false, 1 );
```
westonruter: We might want to bundle these into `theme-plugin-editor.js` instead of adding new scripts.

src/wp-admin/includes/misc.php
```diff
@@ -270,6 +270,180 @@ function update_recently_edited( $file ) {
...
+ * @param int    $level Current level.
+ * @param int    $index Current level.
+ * @param int    $size  Current level.
+ */
+function wp_print_theme_file_tree( $tree, $label = false, $level = 2, $index = 1, $size = 1 ) {
```
westonruter: The `$label` arg is not being used.

westonruter:
Committed in r41851.

PR #282, opened 2017-10-12
Trac #42189: Provide more backward compatibility between MEJS 4.2 and 2.2

rafa8626:
https://core.trac.wordpress.org/ticket/42189


src/wp-includes/js/mediaelement/wp-mediaelement.js
```diff
@@ -3,6 +3,53 @@
...
+		for (var i = 0, total = this.options.features.length; i < total; i++) {
+			var feature = this.options.features[i];
+			if (this['build' + feature]) {
+				try {
+					// Use jQuery for non-default features
+					if (defaultFeatures.indexOf(feature) === -1) {
+						player = $(player);
+						controls = $(controls);
+						layers = $(layers);
+						media = $(media);
```
bradyvercher: Assigning the variables this way causes issues if a custom feature comes before the core features. I also don't think `media` was a jQuery object in previous versions. Something like this may work a little better:

```js
if (defaultFeatures.indexOf(feature) === -1) {
	this['build' + feature](player, $(controls), $(layers), media);
} else {
	this['build' + feature](player, controls, layers, media);
}
```

rafa8626: Gotcha. I totally missed that. I agree with your approach

src/wp-includes/js/mediaelement/wp-mediaelement.js
```diff
@@ -31,14 +31,17 @@
...
 	var init = MediaElementPlayer.prototype.init;
-	MediaElementPlayer.prototype.init = function() {
-		this.$media = this.$node = $(this.node);
+	MediaElementPlayer.prototype.init = function () {
 		this.options.classPrefix = 'mejs-';
+		this.$media = this.$node = $(this.node);
 		init.call(this);
+		this.container = $(this.container);
+		this.controls = $(this.controls);
+		this.layers = $(this.layers);
```
bradyvercher: It looks like these also need to be converted in the `_meReady()` method since it may get called when `MediaElement` is initialized before the `init()` method finishes.  Otherwise, third-party features are built before the elements are converted, which ends up throwing an error.

```js
var ready = MediaElementPlayer.prototype._meReady;
MediaElementPlayer.prototype._meReady = function () {
	this.container = $(this.container);
	this.controls = $(this.controls);
	this.layers = $(this.layers);
	ready.apply(this, arguments);
};
```

rafa8626: Thanks for the note. I'll fix it and re-submit it

src/wp-includes/js/mediaelement/mediaelement-migrate.js
```diff
@@ -1,4 +1,7 @@
-(function (window, $) {
+/* global mejs */
+/* global MediaElementPlayer */
+/* global console */
```
bradyvercher: These should all be on one line to match WP core:

```js
/* global console, MediaElementPlayer, mejs */
```

rafa8626: Done. Thanks for the hint

westonruter:
@rafa8626 I'll have to defer to @bradyvercher's expertise in extending ME.js to know whether it is ready. Once he gives his 👍 then I'll commit it to SVN `trunk` (PRs don't get merged).

rafa8626:
Sounds good thanks for the heads up

bradyvercher:
@westonruter This accounts for the major JS errors I was encountering due to the upgrade. As @rafa8626 mentioned, I think it's critical for the upcoming release. We'd like to do some more testing once it's committed and there are some display issues we'll need to iron out and ship to users before 4.9 is released, so it would be nice to have it land soon if possible.

westonruter:
Committed to trunk in r41877.

PR #284, opened 2017-10-13
Highlight new items button for new menus

brandonpayton:
This is a PR to address:
https://core.trac.wordpress.org/ticket/42114

It adds a reusable function for bouncing a button after a delay unless the user focuses it first. We have [other cases](https://core.trac.wordpress.org/ticket/42211) where this is needed. I'm guessing someone will want to change the name, but `pointOutButton` was the clearest name I came up with.

I've also included work for the following ticket since it involves the same unreviewed reusable function, but it should eventually be split into a separate patch.
https://core.trac.wordpress.org/ticket/42211

### DIFF
--- a/src/wp-admin/js/customize-controls.js
+++ b/src/wp-admin/js/customize-controls.js
@@ -763,10 +763,10 @@
 	 * @param {object} [options]
 	 * @param {number} [options.delay] Delay in milliseconds.
 	 * @param {jQuery} [options.focusTarget] A target for user focus that
-	 *	defaults to the highlighted element. If the user focuses the target
+	 * 	defaults to the highlighted element. If the user focuses the target
 	 * 	before the delay passes, the reminder is canceled. This option exists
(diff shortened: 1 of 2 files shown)


src/wp-admin/js/customize-controls.js
```diff
@@ -751,6 +751,54 @@
...
+	 *
+	 * @param {jQuery} button The element to highlight.
+	 * @param {object} [options]
+	 * @param {number} [options.delay] Delay in milliseconds.
+	 * @param {jQuery} [options.focusTarget] A target for user focus that
+	 * 	defaults to the highlighted element. If the user focuses the target
+	 * 	before the delay passes, the reminder is canceled. This option exists
+	 * 	to accommodate compound buttons containing auxiliary UI,
+	 * 	such as the Publish button augmented with a Settings button.
+	 * @returns {Function} An idempotent function that cancels the reminder.
```
westonruter: I don't see this return function being used anywhere. What is the use case?

PR #285, opened 2017-10-16
Customizer: Add New Menu button to All Locations view

brandonpayton:
This is a PR for addressing [trac-36279](https://core.trac.wordpress.org/ticket/36279).

This adds a "Create New Menu" button per location dropdown to the All Locations panel. Each button expands the New Menu section and selects its location by default.

### DIFF
--- a/src/wp-admin/js/customize-nav-menus.js
+++ b/src/wp-admin/js/customize-nav-menus.js
@@ -1383,6 +1383,26 @@
 
 			// Focus on the new menu section.
 			api.section( customizeId ).focus(); // @todo should we focus on the new menu's control and open the add-items panel? Thinking user flow...
+		},
+
+		/**
+		 * Select a default location.
+		 *
+		 * This method selects a single location by default so we can support
+		 * creating a menu for a specific menu location.
+		 *
+		 * @param {string} locationId - The ID of the location to select.
+		 *  Specifying `null` or `undefined` will clear all selections.
+		 */
+		selectDefaultLocation: function ( locationId ) {
+			var locationControl = api.control( this.id + '[locations]' ),
+				locationSelections = {};
+
+			if ( locationId !== undefined && locationId !== null ) {
+				locationSelections[ locationId ] = true;
+			}
+
+			locationControl.setSelections( locationSelections );
 		}
 	});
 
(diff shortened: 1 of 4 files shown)


src/wp-admin/js/customize-nav-menus.js
```diff
@@ -1383,6 +1383,26 @@
...
 			api.section( customizeId ).focus(); // @todo should we focus on the new menu's control and open the add-items panel? Thinking user flow...
+		},
+
+		/**
+		 * Select a default location.
+		 *
+		 * This method selects a single location by default so we can support
+		 * creating a menu for a specific menu location.
+		 *
+		 * @param {string} locationId - The ID of the location to select.
```
westonruter: Should be `{string|null}` if it accepts both.

src/wp-admin/js/customize-nav-menus.js
```diff
@@ -2349,6 +2376,20 @@
...
 			});
+		},
+
+		/**
+		 * Set the selected locations.
+		 *
+		 * This method sets the selected locations and allows us to do things like
+		 * set the default location for a new menu.
+		 *
+		 * @param {Object.<string,boolean>} selections - A map of location selections.
```
westonruter: Add `@returns {void}` and `@since` tags for this method and the other method.

src/wp-admin/js/customize-nav-menus.js
```diff
@@ -2349,6 +2376,20 @@
...
+		 */
+		setSelections: function ( selections ) {
+			this.container.find( '.menu-location' ).each( function( i, checkboxNode ) {
+				checkboxNode.checked = !! selections[ checkboxNode.dataset.locationId ];
```
westonruter: Probably should use `Boolean()` explicitly, or `! _.isUndefined( selections[ checkboxNode.dataset.locationId ] )`

src/wp-admin/js/customize-nav-menus.js
westonruter: If the value is always going to be `true` then this could just be an array of location IDs, right? I suppose using a hash here is _slightly_ better for performance since we don't have `Set` available yet, as otherwise we'd need to do `indexOf` for each iteration.

src/wp-admin/js/customize-nav-menus.js
brandonpayton: Good point. I fixed the doc and also removed mention of `undefined` as a possible value.

src/wp-admin/js/customize-nav-menus.js
brandonpayton: Fixed. I'll learn...

src/wp-admin/js/customize-nav-menus.js
brandonpayton: I switched to the following expression which is more explicit than what I had.
`locationId in selections ? selections[ locationId ] : false`

brandonpayton:
Thanks for the quick review, @westonruter!

westonruter:
@brandonpayton Please resolve the new conflicts.

westonruter:
Nevermind. It is conflicting because it was committed in r41899 😄

PR #286, opened 2017-10-16
#16841: Capability Queries

swissspidy:
This PR adds capability options to `WP_User_Query`.

It leverages the previous work we've done on querying the capability meta fields, so it doesn't contain too much magic sauce.

Basically, when querying for capabilities, it checks which roles do have these caps and queries for the roles instead. Plus, it queries for the capability as well in case it was added directly to a user.

**To do:**

* [ ] More tests, ideally those that match the ones from [#22212](https://core.trac.wordpress.org/ticket/22212).
* [ ] Add support for these queries in `wp_dropdown_users()`.  
 See [#38135](https://core.trac.wordpress.org/ticket/38135) on Trac for a precedent.
* [ ] Add tests for the current behaviour of `who` queries in `wp_dropdown_users()`.
* [ ] Figure out how/if passing an array for `capability` could work.
* [ ] Multisite: Use `$wp_roles->for_site( $blog_id )` to make sure the roles are fetched from the right site. See [[41625]](https://core.trac.wordpress.org/changeset/41625)
* [ ] Multisite: Add tests accordingly.

See [#16841 on Trac](https://core.trac.wordpress.org/ticket/16841) for more information.


src/wp-includes/class-wp-user-query.php
```diff
@@ -361,6 +372,68 @@ public function prepare_query( $query = array() ) {
...
+		// Capabilities.
+		if ( isset( $qv['capability'] ) || isset( $qv['capability__in'] ) || isset( $qv['capability__not_in'] ) ) {
+			global $wp_roles;
+
+			$available_roles = $wp_roles->roles;
+		}
+
+		$capabilities = array();
+		if ( isset( $qv['capability'] ) ) {
+			// Todo: Add support for arrays? How should these queries be resolved?
```
boonebgorges: The way `role` is handled, vs `role__in`, might be helpful to look at. `role` can be an array, and is handled as an `AND` list: `role=foo,bar` means only users who have *both* the `foo` and `bar` role will be returned. `role__in` is obviously an `IN` / `OR` check. I think it's like this for legacy reasons, though it might be worth emulating for consistency. Additionally, while it's not super useful to be able to match multiple Roles with an `AND`, it's potentially more relevant for caps: show me all users who have `manage_users` AND `manage_options`, or whatever.

felixarntz:
No additonal comments from me except from the multisite support regarding roles possibly need to be switched temporarily before retrieving them.

I like the approach, I definitely agree that thorough tests are needed. Those will uncover possible issues in the logic, but so far I haven't been able to detect any problems. Happy to have another look related to tests in the near future.

swissspidy:
Superseded by https://github.com/WordPress/wordpress-develop/pull/1566

PR #287, opened 2017-10-16
#42220: Transition post/page stubs to drafts when changeset is promoted from auto-draft

westonruter:
https://core.trac.wordpress.org/ticket/42220

- [x] Update publish metabox to display notice about the post/page being part of a changeset, with a link to continue editing changes in Customizer.
- [x] Make sure that drafted stubs still get trashed when a changeset is trashed.
- [x] Add a `display_post_state` for customization drafts.
- [x] Add tests.

### DIFF
--- a/src/wp-admin/includes/meta-boxes.php
+++ b/src/wp-admin/includes/meta-boxes.php
@@ -223,6 +223,26 @@ if ( $can_publish ) : // Contributors don't get to choose the date of publish ?>
 </div><?php // /misc-pub-section ?>
 <?php endif; ?>
 
+<?php if ( 'draft' === $post->post_status && get_post_meta( $post->ID, '_customize_changeset_uuid', true ) ) : ?>
+	<div class="notice notice-info notice-alt inline">
+		<p>
+			<?php
+			echo sprintf(
+				/* translators: %s is the URL to the Customizer */
+				__( 'This is a customization draft and it will be published automatically with its <a href="%s">other customized changes</a>.' ),
+				esc_url(
+					add_query_arg(
+						'changeset_uuid',
+						rawurlencode( get_post_meta( $post->ID, '_customize_changeset_uuid', true ) ),
+						admin_url( 'customize.php' )
+					)
+				)
+			);
+			?>
+		</p>
+	</div>
+<?php endif; ?>
+
 <?php
 /**
  * Fires after the post time/date setting in the Publish meta box.
--- a/src/wp-admin/includes/meta-boxes.php
+++ b/src/wp-admin/includes/meta-boxes.php
@@ -229,7 +229,7 @@ if ( $can_publish ) : // Contributors don't get to choose the date of publish ?>
 			<?php
 			echo sprintf(
 				/* translators: %s is the URL to the Customizer */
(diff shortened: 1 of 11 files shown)


westonruter:
@melchoyce After adding a page stub in the Customizer and then changing the changeset to a draft status, this is then what appears in the post list table:

image

Editing that post will bring it up with a notice in the publish metabox noting that it will be published automatically when the customized changes are published, including a link back to the Customizer to continue making changes or publish them all as a set:

image

You can manually change the status to Private or Publish and in doing so you essentially break the link with the changeset and make it live prematurely.

If you have a drafted set of changes and you have a post/page stub as part of them, when you trash the changeset then the post/page customization draft will be trashed as well. You are able to rescue it out of the Trash still in that case.

Let me know what you think.

I've deployed the changes to http://trac-39896-xwp-testbed.pantheonsite.io/wp-admin/

src/wp-admin/includes/meta-boxes.php
```diff
@@ -223,6 +223,26 @@ function post_submit_meta_box( $post, $args = array() ) {
...
+	<div class="notice notice-info notice-alt inline">
+		<p>
+			<?php
+			echo sprintf(
+				/* translators: %s is the URL to the Customizer */
+				__( 'This is a customization draft and it will be published automatically with its <a href="%s">other customized changes</a>.' ),
+				esc_url(
+					add_query_arg(
+						'changeset_uuid',
+						rawurlencode( get_post_meta( $post->ID, '_customize_changeset_uuid', true ) ),
```
sayedtaqui: Is is important to add the `changeset_uuid` param to the url if branching is not enabled by default?

sayedtaqui:
It would be helpful if we have a edit link in customizer which can directly take the user to relevant page edit screen in a new tab from customizer just as we have a link in edit page to customizer. But may it can be done in future.

melchoyce:
Makes sense to me 👍 

Two minor things:

image

^ Added 10px of margin around the left and right of the notice — having it full-width felt a little weird in this container.

For copy, I'm thinking:

> This draft comes from your [latest customization changes](#). You can edit it, but there's need to publish now. It will be published automatically with those changes.

image

westonruter:
@sayedtaqui I agree it would be ideal to be able to have edit links in the Customizer, but the workflow I think is somewhat confusing if the edit links open in a new window. Also it would be confusing if only the stub posts could have edit links but everything else wouldn't. So I think it would be better to wait until we have a full editing experience in Customizer before we start routing people there explicitly.

src/wp-admin/includes/meta-boxes.php
westonruter: Good question. I think so, because a plugin could enable branching for one user but not for another, and customization drafts would appear together for multiple changesets.

westonruter:
@melchoyce 
> …but there's need to publish now…

I assume this should say that there is _no_ need to publish now?

westonruter:
Also, “latest customization changes” may not be correct if branching is enabled, because there could be multiple changesets together. How about if this says “unpublished customization changes”? This would be in line with the trashing confirmation message which says “unpublished changes”.

westonruter:
image

westonruter:
Committed in r41887.

PR #295, opened 2017-10-27
Restore Customizer new menu control and section

brandonpayton:
This is a PR to address:
https://core.trac.wordpress.org/ticket/42357

The control and section classes for new menus are restored and marked
as deprecated.

The TODO for testing NewMenuControl is also removed due to deprecation.

### DIFF
--- a/src/wp-admin/js/customize-nav-menus.js
+++ b/src/wp-admin/js/customize-nav-menus.js
@@ -3015,7 +3015,7 @@
 	 *
 	 * @constructor
 	 * @augments wp.customize.Control
-	 * @deprecated Since 4.9.0 when it fell out of use due to new menu creation UX.
+	 * @deprecated 4.9.0 This class is no longer used due to new menu creation UX.
 	 */
 	api.Menus.NewMenuControl = api.Control.extend({
 		/**
--- a/src/wp-admin/js/customize-nav-menus.js
+++ b/src/wp-admin/js/customize-nav-menus.js
@@ -1401,7 +1401,7 @@
 
 				if ( checkbox.prop( 'checked' ) ) {
 					navMenuLocationSetting = api( 'nav_menu_locations[' + checkbox.data( 'location-id' ) + ']' );
-					navMenuLocationSetting.set( placeholderId );
+					navMenuLocationSetting.set( menuSection.params.menu_id );
 
 					// Reset state for next new menu
 					checkbox.prop( 'checked', false );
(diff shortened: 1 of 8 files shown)


src/wp-admin/js/customize-nav-menus.js
```diff
@@ -3008,6 +3018,87 @@
...
+			menuSection = api.Menus.createNavMenu( name );
+
+			// Clear name field.
+			nameInput.val( '' );
+			nameInput.removeClass( 'invalid' );
+
+			wp.a11y.speak( api.Menus.data.l10n.menuAdded );
+
+			// Focus on the new menu section.
+			menuSection.focus(); // @todo should we focus on the new menu's control and open the add-items panel? Thinking user flow...
```
celloexpressions: Let's remove this todo, or move it to the appropriate place where it's actually being done. The new flow attempts to get users to click without expanding for them, which may be better in terms of learning for the next time they edit.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -317,13 +317,15 @@ public function __construct( $args = array() ) {
 		require_once( ABSPATH . WPINC . '/customize/class-wp-customize-nav-menu-name-control.php' );
 		require_once( ABSPATH . WPINC . '/customize/class-wp-customize-nav-menu-locations-control.php' );
 		require_once( ABSPATH . WPINC . '/customize/class-wp-customize-nav-menu-auto-add-control.php' );
+		require_once( ABSPATH . WPINC . '/customize/class-wp-customize-new-menu-control.php' ); // @todo Remove in 5.0.
```
celloexpressions: Not sure whether we would typically say that it will be removed in a particular release. Since it's soon, there should be a ticket and patch ready as soon as 4.9 is branched.

src/wp-includes/customize/class-wp-customize-new-menu-control.php
```diff
@@ -5,12 +5,14 @@
  * @package WordPress
  * @subpackage Customize
  * @since 4.4.0
+ * @deprecated 4.9.0 This file is no longer used due to new menu creation UX.
```
celloexpressions: Not sure if we should refer to it as the "new" UX, as it will likely change again eventually. Might be better to reference a specific ticket, or explain why specifically this control is no longer needed.

src/wp-includes/customize/class-wp-customize-new-menu-section.php
```diff
@@ -5,14 +5,18 @@
  * @package WordPress
  * @subpackage Customize
  * @since 4.4.0
+ * @deprecated 4.9.0 This file is no longer used due to new menu creation UX.
```
celloexpressions: Same issue here - let's be more specific about why it's no longer used.

westonruter:
Committed in r42034.

PR #296, opened 2017-10-27
#42360: Customize: Deep-linking to themes section does not work

westonruter:
https://core.trac.wordpress.org/ticket/42360


westonruter:
Committed in r42033.

PR #297, opened 2017-11-01
#42392: Customize: Changeset scheduled month gets unselected if less than 10

westonruter:
https://core.trac.wordpress.org/ticket/42392


westonruter:
Committed in r42104.

PR #298, opened 2017-11-01
#4239: Customize: Starter content does not play nicely with drafted changes

westonruter:
https://core.trac.wordpress.org/ticket/4239


westonruter:
A different approach was taken.

PR #302, opened 2017-11-14
#42548: Widgets: Populate global $post with queried object on singular queries instead of nullifying it

westonruter:
https://core.trac.wordpress.org/ticket/42548


westonruter:
Committed to trunk.

PR #304, opened 2018-01-02
Trac 22249: Add ability to set or remove attributes on enqueued scripts and styles. [WIP]

mehigh:


westonruter:
Looks like a `class.wp-scripts.php.orig` file got committed.

mehigh:
Thanks for noticing, @westonruter 
I've removed it and refreshed the ticket.

